PR #4, opened 2018-07-17
add some attributes to the Options of a Connection

laurentdutheil:
- enableDatabaseCustomization
- brut_force_protection
- import_mode
- disable_signup


alexkappa:
Hi @laurentdutheil, thanks for contributing!

It is a bit strange that these properties are not documented by Auth0. However other users have submitted patches to add properties that are unofficially supported so I can't really say no to such additions :)

Perhaps we can think of a way to support arbitrary values in the `ConnectionOptions` struct, via an embedded `map[string]interface{}` or something along those lines.

Cheers,
Alex

laurentdutheil:
You're so fast!
Thank you!

PR #9, opened 2018-07-30
add CustomScripts and Configuration to Connection Options

laurentdutheil:


alexkappa:
Hi @laurentdutheil, is it possible to find documentation for these fields and their functionality?

The connections resource is getting hard to manage.

Also some examples that I can use to verify this would be awesome

laurentdutheil:
I can explain my usecase.

### Custom scripts
Auth0 supports automatic migration of users from a custom database connection to Auth0.
By activating this feature, your users are moved to Auth0 the first time they log in after you set up the integration. Your users are not asked to reset their password as a result of the migration.
Documentation here: https://auth0.com/docs/users/migrations/automatic

To do this, you have to write custom scripts that connect to the future-old database to retrieve users for your Auth0 Connection

### Configuration values
For Configuration field, I can also explain my need.
When you use multi-tenant in Auth0 (one for production and one for development), your custom scripts do not connect to the same database.
Auth0 supports configuration values that are shared between scripts.
Documentation here: https://auth0.com/docs/dev-lifecycle/setting-up-env#migration

Instead of hard coded values, you can use these configuration values in your scripts to obfuscate passwords or urls. And your scripts are also cross environment compatibles.

### Interest for Auth0 Terraform Provider
When you're provisionning Auth0 Connections with Terraform, it would be great if you can provide these configuration values in your terraform scripts depending of your environment (DEV, PROD...)

NB: For custom scripts, I prefer use [Gitlab Deployment Plugin](https://auth0.com/docs/extensions/gitlab-deploy).
[Ad]You can even read this [article](https://blog.octo.com/en/how-to-do-tdd-on-auth0-scripts/) to facilitate TDD on these custom scripts.[/Ad]

### To improve my contribution
To improve my contribution, do you prefer more tests ? an update of the connection example in the terraform-provider-auth0 project ? anything else ?

alexkappa:
@laurentdutheil thanks for explaining the fields, it really helps me get the idea!

What would be nice is to expand some of our tests to use these new fields, so we can verify they work properly. It would help avoid the typos we had previously 😄 

The TDD approach you linked, I would say although its useful it is out of scope for the purpose this library.

laurentdutheil:
I added an Integration Test for `custom scripts` and `configuration`.

But, to avoid the typos, as Auth0 authorize and persist any option, I don't know how to fail an Integration Test.
So I created a Contract Test that unmarshall a "valid" json. I put quotes because it depends of the test writer. 

I hope this helps.

management/connection_test.go
```diff
@@ -44,6 +99,44 @@ func TestConnection(t *testing.T) {
...
+			t.Fatal("response should contain a custom script with the key 'get_user'")
+		}
+		if value, exist := c.Options.Configuration["KEY_1"]; exist {
+			if value == "VALUE_1" {
+				t.Fatal("response should contain a CRYPTED configuration value for 'KEY_1'")
+			}
+		} else {
+			t.Fatal("response should contain a configuration key 'KEY_1'")
+		}
+		if value, exist := c.Options.Configuration["KEY_2"]; exist {
```
alexkappa: This is pretty much the same test as for KEY_2. Lets skip it.

management/connection_test.go
```diff
@@ -44,6 +99,44 @@ func TestConnection(t *testing.T) {
...
+			"KEY_1": "VALUE_1",
+			"KEY_2": "VALUE_2",
+		}
+
+		err = m.Connection.Update(id, c)
+		if err != nil {
+			t.Error(err)
+		}
+
+		if c.Options.CustomScripts["get_user"] != "function(email, callback) { return callback(null) }" {
```
alexkappa: The js function makes this a bit hard to read, perhaps we can figure out a cleaner way to make this comparison? I was thinking this whole test can be shrinked to something like this:

```
id := c.ID
c.ID = ""       // read-only
c.Name = ""     // read-only
c.Strategy = "" // read-only
c.Options = ConnectionOptions{
  CustomScripts: map[string]interface{}{"get_user": "function(email, callback) { return callback(null) }"},
  Configuration: map[string]interface{}{"foo": "bar"},
}

cc := c // make a copy so we can compare later

err = m.Connection.Update(id, c)
if err != nil {
  t.Error(err)
}

if c.Options.CustomScripts["get_user"] != cc.Options.CustomScripts["get_user"] {
  t.Fatal(`unexpected result for "get_user" custom script`)
}

if _, exist := c.Options.Configuration["foo"]; !exist {
  t.Fatal(`missing key "foo"`)
}
```

laurentdutheil:
I just push the code review.
But I don't understand the failed build.
Can you help me ?

alexkappa:
Hi @laurentdutheil, thanks a lot! I'm not entirely sure either why the build failed, let me try to re-run it in case it was a temporary glitch.

Issue #8, opened 2018-07-23
Improve handling of zero values


alexkappa:
In certain resources we would like to patch with a value that can be considered a zero value in Go terms. Zero values however will be ignored by the `omitempty` struct annotation.

This is a fairly well known issue, described in detail [here](https://willnorris.com/2014/05/go-rest-apis-and-pointers).

The SDKs of popular services such [GitHub](https://github.com/google/go-github) or [Amazon AWS](https://github.com/aws/aws-sdk-go-v2) define the attributes of their objects with pointers, even for scalar values such as integers or strings.

This could be made easier with #6 in place, for backwards compatibility's sake.

PR #10, opened 2018-08-09
improve zero value handling using pointers

alexkappa:
With this PR we can improve the handling of Go's zero values fixing #8 

This is backwards incompatible of course hence the major version update.


hypnoglow:
@alexkappa Do you plan to release v1 only because there are backwards incompatible changes, or because the library can be considered stable enough?

alexkappa:
@hypnoglow there are indeed backwards incompatible changes.

PR #11, opened 2018-08-12
Use log package for dump instead of fmt

hypnoglow:
This allows logging when go-auth0 package is called from
terraform-provider-auth0.

### DIFF
--- a/management/management.go
+++ b/management/management.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"log"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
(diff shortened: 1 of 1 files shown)


alexkappa:
This is great! I was looking for a way to debug from terraform and was having a hard time. Do you happen to have a link describing the use?

hypnoglow:
https://github.com/hashicorp/go-plugin

> Built-in Logging. Any plugins that use the log standard library will have log data automatically sent to the host process. The host process will mirror this output prefixed with the path to the plugin binary. This makes debugging with plugins simple.

I'm not an expert in terraform internals btw, but I think this is the only way to log from plugins.

alexkappa:
This is excellent, thank you! I'll study this a little more.

PR #12, opened 2018-08-12
Introduce AUTH0_DEBUG env variable

hypnoglow:
The variable can be set to enable debug for go-auth0 client. Explicitly
passing WithDebug() option to the client overrides this variable.

### DIFF
--- a/management/management.go
+++ b/management/management.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"net/url"
+	"os"
 	"strconv"
 	"strings"
 	"time"
@@ -129,7 +130,7 @@ func New(domain, clientID, clientSecret string, options ...apiOption) (*Manageme
 		domain:   domain,
 		basePath: "api/v2",
 		timeout:  1 * time.Minute,
-		debug:    false,
+		debug:    debugFromEnv(),
 	}
 
 	for _, option := range options {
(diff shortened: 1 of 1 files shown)


alexkappa:
@hypnoglow what's the use case for this? I feel the `WithDebug()` option would achieve the same outcome.

hypnoglow:
This is also to support your **terraform-provider-auth0**.

The idea is to toggle debugging dynamically. We cannot just hardcode `WithDebug(true)`. Environment variable is a convenient option for this case.

Alternatively, we can handle `AUTH0_DEBUG` env variable in the **terraform-provider-auth0** itself, and pass `WithDebug()` with corresponding value. What would you prefer?

alexkappa:
Toggling it dynamically quite valuable I agree. But I'd rather avoid an env var in the go-auth0 library.

The the more idiomatic way to do this in terraform would be from the [`schema.Provider`](https://github.com/yieldr/terraform-provider-auth0/blob/master/auth0/provider.go#L11-L25) like we do with the rest of the options.

Then the [`configure`](https://github.com/yieldr/terraform-provider-auth0/blob/master/auth0/provider.go#L42-L49) function would know whether to use `WithDebug()` or not.

hypnoglow:
Nice! I'll submit a PR to **terraform-provider-auth0** then.

PR #13, opened 2018-08-12
Add requires_username option to auth0 connection

hypnoglow:
This variable is currently almost undocumented on auth0, but
it works fine for an auth0 database connection.


alexkappa:
@hypnoglow thank you for this! Could you update the test to make use of this field?

alexkappa:
Awesome, thanks @hypnoglow!

PR #14, opened 2018-08-14
Add full definition for Samlp addon

jeff-french:
Here's a first pass at adding support for the complex configuration of the Samlp addon for applications. I don't have much Golang experience so please point out where I've done things wrong.

### DIFF
--- a/management/client.go
+++ b/management/client.go
@@ -65,7 +65,7 @@ type Client struct {
 	CustomLoginPage        string                 `json:"custom_login_page,omitempty"`
 	CustomLoginPagePreview string                 `json:"custom_login_page_preview,omitempty"`
 	FormTemplate           string                 `json:"form_template,omitempty"`
-	Addons                 map[string]interface{} `json:"addons,omitempty"`
+	Addons                 *Addons                `json:"addons,omitempty"`
 
 	// Defines the requested authentication method for the token endpoint.
 	// Possible values are:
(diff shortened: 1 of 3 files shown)


alexkappa:
Hi @jeff-french, really sorry for the late response. This is pretty good, I'll give it some time to review very soon!

management/client.go
```diff
@@ -91,6 +91,37 @@ type ClientJWTConfiguration struct {
 	Algorithm string `json:"alg,omitempty"`
 }
 
+type Addons struct {
+	Aws                  map[string]string `json:"aws,omitempty"`
```
alexkappa: Can we make `go lint` happy by capitalizing abbreviations? For example, this would become AWS, AzureSb would become AzureSB and so on...

management/client.go
```diff
@@ -91,6 +91,37 @@ type ClientJWTConfiguration struct {
...
+	AzureBlob            map[string]string `json:"azure_blob,omitempty"`
+	AzureSb              map[string]string `json:"azure_sb,omitempty"`
+	Rms                  map[string]string `json:"rms,omitempty"`
```
alexkappa: Should we prefer `map[string]interface{}` which leaves us more room in case an addon needs more complex data structure?

management/client.go
```diff
@@ -91,6 +91,37 @@ type ClientJWTConfiguration struct {
...
+	Dropbox              map[string]string `json:"dropbox,omitempty"`
+	EchoSign             map[string]string `json:"echosign,omitempty"`
+	Egnyte               map[string]string `json:"egnyte,omitempty"`
+	Firebase             map[string]string `json:"firebase,omitempty"`
+	NewRelic             map[string]string `json:"newrelic,omitempty"`
+	Office365            map[string]string `json:"office365,omitempty"`
+	Salesforce           map[string]string `json:"salesforce,omitempty"`
+	SalesforceApi        map[string]string `json:"salesforce_api,omitempty"`
+	SalseforceSandboxApi map[string]string `json:"salesforce_sandbox_api,omitempty"`
+	Samlp                *Samlp            `json:"samlp,omitempty"`
```
alexkappa: Same here with capitalization with SalesforceAPI, SalseforceSandboxAPI, SAMLP...

PR #16, opened 2018-08-23
Use oauth2.ReuseToken to re-authenticate when token becomes invalid

### LINKED ISSUE #15: Handle token expiration
Currently, there is a `StaticTokenSource` created that would never refresh the token. This is ok for **terraform-provider-auth0** and similar apps with short-lived sessions.

However, this library cannot be used in long-running applications, like an http server. Once token expires, all requests would fail, and a new `Management` instance have to be created.

Possible solution: replace `StaticTokenSource` with token source implementing token renewal. I'm not sure if this works out-of-the-box for `client_credentials` grant. Also, there is a [https://godoc.org/golang.org/x/oauth2/clientcredentials](https://godoc.org/golang.org/x/oauth2/clientcredentials) package that might provide an easy solution for this.

---

If you approve the idea, I can dig into it to see if I can provide a solution and submit a PR.

alexkappa:
Introduce `oauth2.ReuseToken` in order to refresh our access token for long running processes. Fixes #15 

@hypnoglow would you like to give it a try?

### DIFF
--- a/management/management.go
+++ b/management/management.go
@@ -13,55 +13,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
-
-	"golang.org/x/oauth2"
 )
-
-// Auth embeds a Config and Token structs so it can be used to authenticate our
-// http client.
-type Auth struct {
-	AuthConfig
-	Token
-}
-
-// AuthConfig is the payload used to receive an Auth0 management token. This token
-// is a JWT, it contains specific granted permissions (known as scopes), and it
-// is signed with a application API key and secret for the entire tenant.
-//
-// 	{
-// 	  "audience": "https://YOUR_AUTH0_DOMAIN/api/v2/",
-// 	  "client_id": "YOUR_CLIENT_ID",
-// 	  "client_secret": "YOUR_CLIENT_SECRET",
-// 	  "grant_type": "client_credentials"
-// 	}
-//
-// See: https://auth0.com/docs/api/management/v2/tokens#1-get-a-token
-//
-type AuthConfig struct {
-	Audience     string `json:"audience"`
-	ClientID     string `json:"client_id"`
-	ClientSecret string `json:"client_secret"`
-	GrantType    string `json:"grant_type"`
-}
-
-// Token is the response body from the request to receive an Auth0 management
-// token.
-//
-// 	{
-// 	  "access_token": "eyJ...Ggg",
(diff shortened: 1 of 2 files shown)


hypnoglow:
I've done some testing on this.

The simple script as a test stand:

```go
func main() {
	mgmt, err := management.New("<domain>", "<id>", "<secret>")
	if err != nil {
		panic(err)
	}

	for {
		users, err := mgmt.User.List()
		if err != nil {
			log.Fatal(err)
		}

		log.Printf("got %d users", len(users))

		time.Sleep(time.Second * 5)
	}
}
```

Also, I've set token expiration time for 20 seconds for Management API.

First, I've tried [oauth2/clientcredentials](https://godoc.org/golang.org/x/oauth2/clientcredentials) package and it worked out of the box!
The code I've used: https://github.com/yieldr/go-auth0/compare/master...hypnoglow:test-client-credentials


Results:

```
go run main.go
DEBUG: "get new token"
2018/08/23 23:19:08 got 2 users
2018/08/23 23:19:13 got 2 users
DEBUG: "get new token"
2018/08/23 23:19:18 got 2 users
2018/08/23 23:19:23 got 2 users
DEBUG: "get new token"
2018/08/23 23:19:28 got 2 users
```

However, I think your solution is technically more correct, even if it takes more code.
This is because, strictly speaking, Auth0 documents that requests should be in `application/json`,
rather than `application/x-www-form-urlencoded`. Even if last one, suprisingly, works fine.

So I've also tested your solution:

```
go run main.go
DEBUG: "getting new token 2"
2018/08/23 23:26:32 got 2 users
2018/08/23 23:26:37 got 2 users
DEBUG: "getting new token 2"
2018/08/23 23:26:43 got 2 users
2018/08/23 23:26:48 got 2 users
DEBUG: "getting new token 2"
2018/08/23 23:26:53 got 2 users
2018/08/23 23:26:58 got 2 users
```

Works like a charm!

So I think we should keep your solution, as it provides best effort to comply with
Auth0 documentation.

Thanks!

alexkappa:
Cool, great to see it works! 

I would love to use [oauth2/clientcredentials](https://godoc.org/golang.org/x/oauth2/clientcredentials) as its much more elegant. Thanks for taking a stab at it!

Perhaps I'll get in contact with Auth0 and see what they think about it.

PR #20, opened 2018-11-20
Add support for connection properties ext_nested_groups and max_groups_to_retrieve

mdisibio:
This adds support for 2 more connection properties:  bool ext_nested_groups, string max_groups_to_retrieve.  max_groups_to_retrieve is handled as a string in the Auth0 API, but its content is numeric.  Not sure if there is a better way to handle this.


alexkappa:
Thank you @mdisibio!

PR #21, opened 2018-11-25
signing_keys JSON

harryrose:
While extending the yieldr auth0 terraform provider to add a new client data resource, I noticed that the auth0 client was ignoring the signing_keys part of the client response. This PR now respects signing_keys in the Auth0 response.

The Auth0 API insists that the field isn't present on submission, so I've written a custom marshaller which makes a copy of the structure and sets the signing_keys field to nil before marshalling.

I was in two (or more...) minds on how to do this.  The usual approach seems to be to duplicate the structure with the necessary json tags, but for such a large structure, that would be a pain to maintain.

Another approach could be to have two structs, one of which omits the field entirely, but that would constitute a breaking change:

```
type ClientPost struct {
 ... all fields except SigningKeys
}

type Client struct {
  SigningKeys []map...
  ClientPost
}
```

### DIFF
--- a/management/client.go
+++ b/management/client.go
@@ -46,7 +46,7 @@ type Client struct {
 	JWTConfiguration  *ClientJWTConfiguration `json:"jwt_configuration,omitempty"`
 
 	// Client signing keys
-	SigningKeys   []map[string]string `json:"-"`
+	SigningKeys   []map[string]string `json:"signing_keys,omitempty"`
 	EncryptionKey map[string]string   `json:"encryption_key,omitempty"`
 	SSO           *bool               `json:"sso,omitempty"`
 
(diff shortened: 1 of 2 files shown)


management/client.go
```diff
@@ -84,6 +84,14 @@ func (c *Client) String() string {
...
 }
 
+// MarshalJSON encodes the client structure to JSON excluding the signing_keys field
+func (c *Client) MarshalJSON()([]byte, error) {
+	copy := *c
+	copy.SigningKeys = nil
+	// Marshal copy here, not &copy, to avoid recursion
+	return json.Marshal(copy)
+}
+
```
alexkappa: @harryrose why do this in particular for signing keys? There are many other fields that are read-only, but we don't actively omit them during serialization but rather leave it up to users to do so.

Could we omit the `MarshalJSON` and let users set `SigningKeys` to nil when creating/updating?

PR #25, opened 2018-12-09
Add is_token_endpoint_ip_header_trusted property support to Client

ar3cka:
This PR introduces ability to change is_token_endpoint_ip_header_trusted property to clients.

There is no documentation regarding the property, but the api returns and allows to change the value of it.

p.s. I've been told by auth0 support that they will update the documentation


alexkappa:
Thanks @ar3cka!

PR #33, opened 2019-03-06
Add support for verification email jobs

SimonBarendse:
Add support for this endpoint: https://auth0.com/docs/api/management/v2#!/Jobs/post_verification_email

The function uses a different struct for the response than is used for the request body. What do you think about that?


alexkappa:
Hi @SimonBarendse, thank you for giving this a go!

In general I'd prefer to keep the `request` API without `in` and `out` parameters in order to maintain the style used in other resources. 

We could achieve this by combining the properties of both request and response structs. As I see a lot of similarity with other jobs such as exporting and importing, we could extend the `Job` struct with the fields of those too. I have the feeling that creating specific structs for each job would pollute the API unnecessarily.

With that in mind, a job would look something like this:

```Go
type Job struct {
  ID *string `json:"id,omitempty"`
  Type *string `json:"type,omitempty"`
  Status *string `json:"status,omitempty"`
  CreatedAt *string `json:"created_at,omitempty"`
  UserID *string `json:"user_id,omitempty"`
  ClientID *string `json:"cliend_id,omitempty"`
  ConnectionID *string `json:"connection_id,omitempty"`
  ...
}
```
and a `JobManager` would use it as:

```Go
func (jm *JobManager) VerifyEmail(j *Job) error {
	return jm.m.post(jm.m.uri("jobs/verification-email"), j)
}
```

And using it would look like:

```Go
api.Job.VerifyEmail(&Job{UserID: "..."})
```

SimonBarendse:
Hi @alexkappa, thanks for your quick response!

I feel like specific structs would make it more intuitive what can be expected to be present after return. However, as you also point out, this is a different style than we use for the other endpoints.

I'll change my PR to use a single struct to be consistent with the style we use for the other endpoints.

management/job.go
```diff
@@ -0,0 +1,25 @@
...
+
+type JobManager struct {
+	m *Management
+}
+
+func NewJobManager(m *Management) *JobManager {
+	return &JobManager{m}
+}
+
+func (jm *JobManager) VerificationEmail(j *Job) error {
```
alexkappa: Could we turn this into an action (e.g. `VerifyEmail`) to be aligned with other functions such as `Read`, `Update` etc?

management/job.go
```diff
@@ -0,0 +1,25 @@
...
+import "time"
+
+type Job struct {
+	ID        *string    `json:"id,omitempty"`
+	Status    *string    `json:"status,omitempty"`
+	Type      *string    `json:"type,omitempty"`
+	CreatedAt *time.Time `json:"created_at,omitempty"`
+
+	UserID   *string `json:"user_id,omitempty"`
+	ClientID *string `json:"cliend_id,omitempty"`
```
alexkappa: Can you add comments from the schema description in the management api docs?

For example 

```Go
// The job's identifier. Useful to retrieve its status
ID *string `json:"id,omitempty"`
```

management/job_test.go
```diff
@@ -0,0 +1,42 @@
...
+	defer m.User.Delete(userID)
+
+	t.Run("verification email", func(t *testing.T) {
+		job := &Job{
+			UserID: auth0.String(userID),
+		}
+
+		err = m.Job.VerificationEmail(
+			job,
+		)
```
alexkappa: Can we condense this into a single line?

SimonBarendse:
Hi @alexkappa , thanks for your feedback! I have processed it.

management/job_test.go
```diff
@@ -0,0 +1,40 @@
...
+	err = m.User.Create(u)
+	if err != nil {
+		t.Error(err)
+	}
+
+	userID := auth0.StringValue(u.ID)
+
+	defer m.User.Delete(userID)
+
+	t.Run("verification email", func(t *testing.T) {
```
alexkappa: Can we rename this subtest? `t.Run("VerifyEmail", ..`. Last thing and we're good to merge ;)

SimonBarendse: Sure :+1: It's good to be thorough :smile:

alexkappa:
Nice work @SimonBarendse, thank you so much for working on this!

SimonBarendse:
@alexkappa No problemo! Thank you for staying on top of this and responding so quickly :smile:

PR #37, opened 2019-03-27
Support for Roles

### LINKED ISSUE #35: Auth0 add roles to API v2
Auth0 has added their authorization model using roles and permissions to their API:
  https://auth0.com/docs/api/management/v2#!/Roles/get_roles
  https://auth0.com/docs/api/management/v2#!/Users/get_roles

I would really love to work on this but I'm new to golang and therefore would need a little assistance.

alexkappa:
Resolves #35.

Added `Role` and `Permission` entities as well as a `RoleManager` which supports the following functionality:

- `Create`: Creates a role
- `Read`: Reads a role
- `Update`: Updates a role
- `List`: Lists all available roles within the current tenant
- `AssignUsers`: Assigns a user to a role
- `Users`: Lists all the users associated to this role.
- `AssignPermissions`: Assigns a permission to a role
- `Permissions`: Lists all the permissions associated to this role.
- `UnassignPermissions`: Unassigns a permission from a role
- `Delete`: Deletes a role

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ Following is a list of supported Auth0 resources.
 - [x] [Grants](https://auth0.com/docs/api/management/v2#!/Grants/get_grants)
 - [x] [Logs](https://auth0.com/docs/api/management/v2#!/Logs/get_logs)
 - [x] [Resource Servers (APIs)](https://auth0.com/docs/api/management/v2#!/Resource_Servers/get_resource_servers)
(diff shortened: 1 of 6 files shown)


ltiessen:
It would be great if we could get this reviewed and merged. I'm looking forward to being able to stop using the Auth0 Authorization Extension and having Terraform build these new roles during deploy instead. In order to do that though we need this functionality.

PR #38, opened 2019-04-03
Support roles to users

sinnedh:
This PR is based on branch `roles`.

Adding functionality to User
 - GetRoles
 - AssignRole
 - UnassignRoles

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,6 @@ Following is a list of supported Auth0 resources.
 - [x] [Resource Servers (APIs)](https://auth0.com/docs/api/management/v2#!/Resource_Servers/get_resource_servers)
 - [x] [Roles](https://auth0.com/docs/api/management/v2#!/Roles)
 - [x] [Rules](https://auth0.com/docs/api/management/v2#!/Rules/get_rules)
-- [ ] [Roles](https://auth0.com/docs/api/management/v2#!/Roles/get_roles)
 - [x] [Rules Configs](https://auth0.com/docs/api/management/v2#!/Rules_Configs/get_rules_configs)
 - [ ] [User Blocks](https://auth0.com/docs/api/management/v2#!/User_Blocks/get_user_blocks)
 - [x] [Users](https://auth0.com/docs/api/management/v2#!/Users/get_users)
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ Following is a list of supported Auth0 resources.
 - [x] [Resource Servers (APIs)](https://auth0.com/docs/api/management/v2#!/Resource_Servers/get_resource_servers)
 - [x] [Roles](https://auth0.com/docs/api/management/v2#!/Roles)
 - [x] [Rules](https://auth0.com/docs/api/management/v2#!/Rules/get_rules)
+- [ ] [Roles](https://auth0.com/docs/api/management/v2#!/Roles/get_roles)
 - [x] [Rules Configs](https://auth0.com/docs/api/management/v2#!/Rules_Configs/get_rules_configs)
 - [ ] [User Blocks](https://auth0.com/docs/api/management/v2#!/User_Blocks/get_user_blocks)
 - [x] [Users](https://auth0.com/docs/api/management/v2#!/Users/get_users)
--- a/management/user_test.go
+++ b/management/user_test.go
@@ -28,17 +28,17 @@ func TestUser(t *testing.T) {
 		},
 	}
 
-    r1 := &Role{
-	    Name:        auth0.String("admin"),
-	    Description: auth0.String("Administrator"),
+	r1 := &Role{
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -52,6 +52,7 @@ Following is a list of supported Auth0 resources.
 - [x] [Resource Servers (APIs)](https://auth0.com/docs/api/management/v2#!/Resource_Servers/get_resource_servers)
 - [x] [Roles](https://auth0.com/docs/api/management/v2#!/Roles)
 - [x] [Rules](https://auth0.com/docs/api/management/v2#!/Rules/get_rules)
+- [ ] [Roles](https://auth0.com/docs/api/management/v2#!/Roles/get_roles)
```
alexkappa: Roles is defined just above. We should remove one of them.

management/user_test.go
```diff
@@ -80,11 +92,35 @@ func TestUser(t *testing.T) {
...
 		}
+		t.Logf("%v\n", roles)
+	})
+
+	t.Run("AssignRoles", func(t *testing.T) {
+		roles := []*Role{r1, r2}
+		err = m.User.AssignRoles(auth0.StringValue(u.ID), roles...)
+		if err != nil {
+			t.Error(err)
+			t.Fatal(err)
```
alexkappa: I think you need just one. Maybe stick to `t.Error`? Not sure why but in most tests I used `Error` unless without the test the following tests wouldn't be useful. Usually when followed by a `DeleteXXX` I avoid fatal errors as they wouldn't run and leave orphaned resources created in the auth0 account.

management/user_test.go
```diff
@@ -132,4 +168,8 @@ func TestUser(t *testing.T) {
 			m.User.Delete(auth0.StringValue(user.ID))
 		}
 	})
+
+	m.Role.Delete(auth0.StringValue(r1.ID))
+	m.Role.Delete(auth0.StringValue(r2.ID))
```
alexkappa: Perhaps move this close to the creation using `defer`? Just to keep the consistency with other tests.

management/user_test.go
```diff
@@ -28,6 +28,18 @@ func TestUser(t *testing.T) {
 		},
 	}
 
+	r1 := &Role{
+		Name:        auth0.String("admin"),
+		Description: auth0.String("Administrator"),
+	}
+	m.Role.Create(r1)
```
alexkappa: Good to test for errors here.

sinnedh:
@alexkappa Thanks a lot for your feedback. I have fixed the issues that you have mentioned.

management/user_test.go
```diff
@@ -80,11 +101,34 @@ func TestUser(t *testing.T) {
...
 			t.Error(err)
 		}
+
+	})
+
+	t.Run("UnassignRoles", func(t *testing.T) {
+		roles := []*Role{r1, r2}
+		err = m.User.UnassignRoles(auth0.StringValue(u.ID), roles...)
+		if err != nil {
+			t.Fatal(err)
```
alexkappa: Can we make this a `t.Error()` please? Then we're good to merge ;)

sinnedh: Sorry, I misunderstood your previous comment... no its fixed.

alexkappa: Uhm, did you forget to commit the change? 😄

management/job_test.go
```diff
@@ -1,88 +0,0 @@
-package management
```
alexkappa: Why did you delete this file?

management/user_test.go
sinnedh: Argh, I have committed the changes to the wrong file (I removed `job_test.go` locally for testing). I have undone these changes now.

management/user_test.go
```diff
@@ -80,13 +101,36 @@ func TestUser(t *testing.T) {
...
+		roles, err = m.User.GetRoles(auth0.StringValue(u.ID))
+		if err != nil {
+			t.Error(err)
+		}
+		t.Logf("%v\n", roles)
+	})
+
+	t.Run("AssignRoles", func(t *testing.T) {
+		roles := []*Role{r1, r2}
+		err = m.User.AssignRoles(auth0.StringValue(u.ID), roles...)
```
alexkappa: This can also be written using variadic args:

```
m.User.AssignRoles(auth0.StringValue(u.ID), r1, r2)
```

management/user_test.go
alexkappa: Ok. In case you want to run tests in isolation, what I usually do is run a specific test case using the `-run` args to `go test`.

For example running the user tests you could do the following:

```
go test -v ./... -run ^TestUser$
```

management/user_test.go
sinnedh: Nice. Thanks!

PR #50, opened 2019-06-04
Added new options to allow custom connection type configurations

jordan-simonovski:
Adding in some more configuration items for Auth0 options to allow custom connection types, which in this case are OAuth2 related.
@alexkappa I'll be raising another PR for the terraform provider after this one :) Let me know if there's anything I'm missing.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module gopkg.in/auth0.v1
+module github.com/jordan-simonovski/auth0
 
 go 1.12
 
--- a/management/connection.go
+++ b/management/connection.go
@@ -117,6 +117,10 @@ type ConnectionOptions struct {
 
 	// Adfs
 	AdfsServer *string `json:"adfs_server,omitempty"`
+
+	// Options to add Custom Authorization type for OAuth2 Connections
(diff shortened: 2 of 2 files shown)


jordan-simonovski:
Hey @alexkappa! would you have a chance to look a this soon? 😄

alexkappa:
Hi @jordan-simonovski, could you add a test for this? I don't have any documentation at hand so its a little hard to verify. A test demonstrating its use would be very helpful

jordan-simonovski:
Hey @alexkappa I'm not sure what kind of test to add to this one, since these are just additional options to the existing set of options for the connection creation that already seems to have tests. I've got tests on the provider repo itself, but for this one I was just adding more options to the json string.

go.mod
```diff
@@ -1,4 +1,4 @@
-module gopkg.in/auth0.v1
+module github.com/jordan-simonovski/auth0
```
alexkappa: This change needs to be reverted

alexkappa:
@jordan-simonovski perhaps set some of the newly introduced fields in the existing tests to verify whether they are working?

PR #51, opened 2019-06-19
Add missing fields in Tenant

laurentdutheil:
- idle_session_lifetime
- universal_login
- enabled_locales
- device_flow

and flags :
- enable_sso
- allow_changing_enable_sso
- universal_login
- enable_legacy_logs_search_v2
- disable_clickjack_protection_headers

### DIFF
--- a/management/tenant.go
+++ b/management/tenant.go
@@ -45,6 +45,17 @@ type Tenant struct {
 
 	// A set of available sandbox versions for the extensibility environment
 	SandboxVersionAvailable []interface{} `json:"sandbox_versions_available,omitempty"`
+
+	// Force a user to login after they have been inactive for the specified number (unit: hours)
+	IdleSessionLifetime *int `json:"idle_session_lifetime,omitempty"`
+
+	// Used to store additional metadata
+	UniversalLogin *TenantUniversalLogin `json:"universal_login,omitempty"`
+
+	// Supported locales for the UI
+	EnabledLocales []interface{} `json:"enabled_locales,omitempty"`
+
+	DeviceFlow *TenantDeviceFlow `json:"device_flow,omitempty"`
 }
 
 func (t *Tenant) String() string {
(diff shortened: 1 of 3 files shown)


gtirloni:
It'd be awesome if we could have this merged.

management/tenant_test.go
```diff
@@ -11,6 +13,83 @@ func TestTenant(t *testing.T) {
 	var tn *Tenant
 	var err error
 
+	// Based on Auth0 examples, we can verify that the definition of the struct is OK
+	// by unit testing a parse of a json
+	t.Run("should parse all fields", func(t *testing.T) {
```
alexkappa: I don't think this is the best way to verify a resource. It verifies json decoding but thats about it. Perhaps we could enable some fields in our tenant and assert whether they are set or not?

Additionally, are these fields documented by auth0? If so it could help me verify.

management/tenant.go
```diff
@@ -105,6 +116,42 @@ type TenantFlags struct {
 	// If enabled, All your email links and urls will use your configured custom
 	// domain. If no custom domain is found the email operation will fail.
 	EnableCustomDomainInEmails *bool `json:"enable_custom_domain_in_emails,omitempty"`
+
+	// If enabled, users will not be prompted to confirm log in before SSO redirection.
+	EnableSso *bool `json:"enable_sso,omitempty"`
```
alexkappa: Can we name this `EnableSSO` to more inline with Go best practices to capitalise abbreviations?

management/tenant_test.go
laurentdutheil: My intention was to validate the json "schema" with a unit test to avoid typos.
I based the name of the fields with this documentation : https://auth0.com/docs/api/management/v2#!/Tenants/get_settings

I think it would be difficult to validate all the fields with an integration test because some fields may be incompatible with each oher.

I also think this simple test can be used as a documentation/exemple for the resource.


But, having said that, if you do not agree, I prefer to align myself with the community's opinion by removing this test.
What do you think ?

management/tenant.go
laurentdutheil: yes !

management/tenant_test.go
alexkappa: I understand, it's also not so easy to test all these options as the tenant resource is not one which we can create/delete on demand. 

I'd still prefer to keep things aligned with the rest of the code and perhaps do these types of tests as unit tests in the future.

alexkappa:
@laurentdutheil I believe the tests are failing due to our default connection requiring usernames now. This was done in order to test the username and nickname in the users resource (see #56).

Is it possible to rebase off the latest master and see if the issue is fixed?

management/tenant_test.go
laurentdutheil: I totally ok with that, but...
...I created the same tests for the Branding resource and the PR has just been merged (sorry).

I propose:
1/ I suppress the tests both in tenant_test and branding_test
2/ I add the same tests for all the resources (in antoher PR)
3/ I add the same tests for all the resources in another file (ex: tenant_u_test) (also in antoher PR)

Choose and I can do it on monday.

laurentdutheil: I choose 1/ 
And I'll propose another PR to de 3/

alexkappa:
Thanks for this @laurentdutheil!

PR #54, opened 2019-06-28
Add enforce_policies field to resource server

23doors:
https://auth0.com/docs/api/management/v2#!/Resource_Servers/patch_resource_servers_by_id


23doors:
tests are failing but it's not a related issue to this PR.

23doors:
@alexkappa could you take a look when you have a chance? we're using auth0 TF provider and want to add PR there to map this field.

alexkappa:
Hi @23doors I'll have a look into why the tests are failing, but this seems good to go. I suspect its related to me changing the tests to use my own auth0 account which doesn't have permissions to create custom domains.

PR #56, opened 2019-07-10
User add family and given name

maier-stefan:
As a user i want to have access to the full user profile.

Reading https://auth0.com/docs/users/references/user-profile-structure we should be able to retrieve given_name and family_name.

### DIFF
--- a/management/user.go
+++ b/management/user.go
@@ -26,6 +26,12 @@ type User struct {
 	// The users name
 	Name *string `json:"name,omitempty"`
 
+	// The users given name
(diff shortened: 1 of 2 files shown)


alexkappa:
Thanks for submitting this @maier-stefan!

maier-stefan:
@alexkappa can you please make a release (v1.2.4) so i can pick the latest changes via go modules. Thanks a lot

alexkappa:
Sure thing, coming up

Issue #30, opened 2019-01-21
Return an error message from malformed requests


alexkappa:
When creating the requests to send to the Auth0 API, we don't check for errors coming from [`http.NewRequest(method, uri, &payload)`](https://github.com/yieldr/go-auth0/blob/master/management/management.go#L135-L136).

If the creation of the client was done with a malformed URL, this would fail without a helpful error message.

It could also be good to validate the URL when creating a new client with `management.New()`.

This issue was reported in yieldr/terraform-provider-auth0/issues/74

PR #59, opened 2019-07-11
Fix domain related error

bogdanpetrea:
This fixes a silent error we were getting when passing the auth0 domain with included web protocol (`https://tenant.auth0.com`)

Changes:
- The code should be now more resilient to this case, but it should still work with the case where the protocol is not specified.
- If an error arises when making a request to auth0, it is no longer silently ignored.

### DIFF
--- a/management/management.go
+++ b/management/management.go
@@ -74,7 +74,7 @@ type Management struct {
 	// Branding settings such as company logo or primary color.
 	Branding *BrandingManager
 
-	domain   string
+	url      *url.URL
 	basePath string
 	timeout  time.Duration
 	debug    bool
@@ -85,9 +85,17 @@ type Management struct {
 // New creates a new Auth0 Management client by authenticating using the
 // supplied client id and secret.
 func New(domain, clientID, clientSecret string, options ...apiOption) (*Management, error) {
+	if !strings.HasPrefix(domain, "http://") && !strings.HasPrefix(domain, "https://") {
+		domain = "https://" + domain
+	}
+
+	u, err := url.Parse(domain)
+	if err != nil {
+		return nil, err
+	}
 
 	m := &Management{
-		domain:   domain,
+		url:      u,
 		basePath: "api/v2",
 		timeout:  1 * time.Minute,
 		debug:    false,
@@ -97,7 +105,7 @@ func New(domain, clientID, clientSecret string, options ...apiOption) (*Manageme
 		option(m)
 	}
 
-	m.http = client.OAuth2(domain, clientID, clientSecret)
+	m.http = client.OAuth2(m.url, clientID, clientSecret)
 	m.http = client.WrapUserAgent(m.http)
 	m.http = client.WrapRetry(m.http)
 	if m.debug {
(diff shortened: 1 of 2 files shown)


management/management.go
```diff
@@ -142,12 +150,16 @@ func (m *Management) q(options []reqOption) string {
...
 	var payload bytes.Buffer
 	if v != nil {
 		json.NewEncoder(&payload).Encode(v)
 	}
-	req, _ := http.NewRequest(method, uri, &payload)
+
+	req, err := http.NewRequest(method, uri, &payload)
+	if err != nil {
+		return err
+	}
```
laurentdutheil: That will fix #30

management/management.go
```diff
@@ -71,7 +71,7 @@ type Management struct {
 	// Stat is used to retrieve usage statistics.
 	Stat *StatManager
 
-	domain   string
+	domain   *url.URL
```
alexkappa: Can we rename `domain` to `url` then since its holding a URL now?

management/management.go
```diff
@@ -82,9 +82,17 @@ type Management struct {
 // New creates a new Auth0 Management client by authenticating using the
 // supplied client id and secret.
 func New(domain, clientID, clientSecret string, options ...apiOption) (*Management, error) {
+	parsedDomain, err := url.Parse(domain)
```
alexkappa: If we rename to url then this would make more sense being `u, err := url.Parse(domain)`. Do you agree?

alexkappa:
The tests are failing due to a nil pointer error, any idea why?

bogdanpetrea:
Thanks for reviewing @alexkappa!
I think the panic is because `JWTConfiguration` is actually nil [here](https://github.com/go-auth0/auth0/blob/master/management/client_test.go#L50). I don't know how exactly my changes brought up this error though.

And then some tests are failing because the domain url is somehow not parsed properly, and then a POST request is being made to an incomplete url `https://oauth/token`. I'll see what I can find by printing the domain at various stages in the code.

bogdanpetrea:
Turns out the problem was related to this [stackoverflow question](https://stackoverflow.com/questions/46719948/how-to-add-scheme-and-path-to-url-url) (tl;dr: when the scheme is not specified, the host is parsed as the path). I've met a similar situation with Python's [furl](https://github.com/gruns/furl) and I still didn't expect it :neutral_face:.

management/management.go
```diff
@@ -85,9 +85,17 @@ type Management struct {
...
 // supplied client id and secret.
 func New(domain, clientID, clientSecret string, options ...apiOption) (*Management, error) {
+	if !strings.HasPrefix(domain, "http://") && !strings.HasPrefix(domain, "https://") {
+		domain = "https://" + domain
+	}
+
+	u, err := url.Parse(domain)
+	if err != nil {
+		return nil, err
+	}
```
alexkappa: This is great, so much better! From the playground link you posted I got this idea, what if we check for the presense of `"//"` instead, so we can account for any scheme, and replace it with `https://`.

```
if i := strings.Index(s, "//"); i != -1 {
	s = s[i+2:]
}
s = "https://" + s

u, err := url.Parse(domain)
if err != nil {
	return nil, err
}

if u.Host == "" {
	return errors.New("Domain should not be blank")
}
```
Link to [playground](https://play.golang.org/p/5WwZifw_ikQ).

alexkappa:
Merged this in, thank you @bogdanpetrea!

bogdanpetrea:
Thanks! And sorry because I got caught with other stuff and I kind of forgot about this PR.

PR #60, opened 2019-07-16
Stabilize job test

laurentdutheil:
The connectionID is hardcoded in Job test.
This fails when another contributor want to execute this on his own tenant.


alexkappa:
Thanks @laurentdutheil, this is nice. Could you rebase so you have the latest changes so the tests can pass? Its good to merge as soon as tests are green

PR #62, opened 2019-07-23
add unit tests to validate the json "schemas" to avoid typos

laurentdutheil:
Also think this simple tests can be used as a documentation/example for the resources.

### DIFF
--- a/management/branding_u_test.go
+++ b/management/branding_u_test.go
@@ -1,0 +1,49 @@
+package management
+
+import (
+	"encoding/json"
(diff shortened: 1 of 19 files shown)


alexkappa:
Hi @laurentdutheil, I appreciate you took the time to write this up. But I'm afraid this adds more work for me as a maintainer and would prefer to avoid merging it in.

laurentdutheil:
I totaly aggre. Thank you for your concern.

PR #64, opened 2019-08-02
fix: check error when creating http.Request

RichardKnop:
Check error when creating new http.Request. Otherwise, when terraform apply command fails because of an issue with creating of request, you will get panic instead of nice error. This will instead print actual error so you can see what happened.

### DIFF
--- a/management/job.go
+++ b/management/job.go
@@ -120,7 +120,10 @@ func (jm *JobManager) ImportUsers(j *Job) error {
 	}
 	mp.Close()
 
-	req, _ := http.NewRequest("POST", jm.m.uri("jobs/users-imports"), &payload)
+	req, err := http.NewRequest("POST", jm.m.uri("jobs/users-imports"), &payload)
+	if err != nil {
+		return err
+	}
 	req.Header.Add("Content-Type", mp.FormDataContentType())
 
 	ctx, cancel := context.WithTimeout(context.Background(), jm.m.timeout)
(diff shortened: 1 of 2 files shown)


RichardKnop:
Here is nice error after this fix:

```
auth0_client.tenant_application: Creating...

Error: parse https://https:%2F%2Fexample.com/api/v2/clients: invalid URL escape "%2F"

  on application.tf line 1, in resource "auth0_client" "tenant_application":
   1: resource "auth0_client" "tenant_application" {
```

Instead of ugly error before:

```
auth0_client.tenant_application: Creating...

Error: rpc error: code = Unavailable desc = transport is closing


panic: runtime error: invalid memory address or nil pointer dereference
2019-08-02T12:35:54.317+0100 [DEBUG] plugin.terraform-provider-auth0: [signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x1989869]
2019-08-02T12:35:54.317+0100 [DEBUG] plugin.terraform-provider-auth0: 
2019-08-02T12:35:54.317+0100 [DEBUG] plugin.terraform-provider-auth0: goroutine 25 [running]:
2019-08-02T12:35:54.317+0100 [DEBUG] plugin.terraform-provider-auth0: gopkg.in/auth0.v1/management.(*Management).request(0xc0005240c0, 0x1c68245, 0x4, 0xc0000d6780, 0x3e, 0x1b15d80, 0xc00052a600, 0x0, 0x0)
```

alexkappa:
Thank you @RichardKnop ;)

PR #67, opened 2019-08-28
Add List method to ClientGrantManager

eclipse1985:


alexkappa:
Hi @eclipse1985 apologies for the late reply. Thanks for the submission! Any chance we can add a test to this?

alexkappa:
Merged in a previous PR that does the same.

PR #85, opened 2019-12-24
Make reqOption public to allow users of the library to mock it

orishoshan:
Currently, the library can _almost_ be mocked for testing: all of the concrete types used in its interface are data structures, with the exception of reqOption.

### DIFF
--- a/management/management.go
+++ b/management/management.go
@@ -167,7 +167,7 @@ func (m *Management) uri(path ...string) string {
 	}).String()
 }
 
-func (m *Management) q(options []reqOption) string {
+func (m *Management) q(options []ReqOption) string {
 	if len(options) == 0 {
 		return ""
 	}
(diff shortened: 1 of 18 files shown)


orishoshan:
This PR had previously accidentally included some commits I hadn't meant to put into it that are unrelated to the change in the title - it now includes _only_ the `reqOption` change.

alexkappa:
Hi @orishoshan, thanks for contributing! I understand the use case of mocking out services you include in your own codebase. For what it's worth I made it private on purpose so one would rely on the existing methods for setting options in requests.

I think the changes you propose in this PR are ok, but let me take a better look and get back to you soon.

management/management.go
```diff
@@ -299,20 +299,20 @@ func (m *managementError) Status() int {
 	return m.StatusCode
 }
 
-// reqOption configures a call (typically to retrieve a resource) to Auth0 with
+// ReqOption configures a call (typically to retrieve a resource) to Auth0 with
 // query parameters.
-type reqOption func(url.Values)
+type ReqOption func(url.Values)
```
alexkappa: Can we use the name `RequestOption` for this instead? As its becoming part of the public API maybe its okay to be more verbose.

alexkappa:
With v3 this issue is addressed together with support for totals and other improvements.

orishoshan:
Hi @alexkappa,

Sorry for not responding! I forgot I had turned off email notifications for github. The changes made in v3 indeed address this, thank you.

Issue #63, opened 2019-08-01
Failures when using IncludeTotals(true)


mdobosz-isp:
When using the [IncludeTotals](https://github.com/go-auth0/auth0/blob/2b97815603cd8b6e18ae3a09889edb9bd93a94e2/management/management.go#L287) request option with something like the Users endpoint, a json unmarshal error occurs because the response body shape changes from an array of users to an object containing a field with the array of users.

```go
users, err := a.m.User.Search(
	management.IncludeTotals(true),
	management.Parameter("sort", "email:1"),
)
// err is "json: cannot unmarshal object into Go value of type []*management.User"
```

Before I start submitting PRs I wanted to validate that this is something you are interested in taking in.

PR #87, opened 2020-01-07
Add support for searching for users with totals

mdobosz-isp:
Fixes go-auth0/auth0#63


alexkappa:
Hi @mdobosz-isp thanks for giving this a go. It is indeed something that we need to improve, but with `...WithTotals` methods will increase the surface of the api quite significantly.

My intention with `reqOption` was to allow users to easily customize the response. Where this doesn't work is when the actual request changes if you use the `include_totals` parameter.

I'm thinking that the best way to go is something akin to the [google/go-github](/google/go-github) package. Where they return three values: a slice of elements (e.g. users), a list instance and an error.

```
func (um *UserManager) List(opts ...reqOption) (u []*User, l *List, err error) {
    // ...
}
```

This would be a backwards incompatible change though, se we'll need to do this in a new major version.

alexkappa:
Another option would be to change `List` or `Search` methods to return `XXXList` instead of a `[]*XXX` which is more in line with your approach here.

We could then force `include_totals` on these methods.

alexkappa:
Hi @mdobosz-isp, I have added support for including totals in a new version (v3.0.0). Feel free to give it a go. As I've done many improvements I've been meaning to get to with this change. Thanks again for inspiring the way this feature came through.

PR #93, opened 2020-02-04
bugfix: user identity field user_id is a number in some id providers

zeevmoney:
When the auth0 identity provider is Github, the user field might a number and not a "string"
This causes the Unmarshal JSON() func to fail with error:

```cannot unmarshal number into Go struct field UserIdentity.user_id of type string```

example:
```json
{
            "provider": "github",
            "user_id": 123123123,
            "connection": "github",
            "isSocial": true
}
```

### DIFF
--- a/management/user.go
+++ b/management/user.go
@@ -1,6 +1,11 @@
 package management
 
-import "time"
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"time"
+)
 
 // User represents an Auth0 user resource
 //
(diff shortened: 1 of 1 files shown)


alexkappa:
Hi @zeevmoney, thanks for pointing out the issue and for working on a fix. It is perhaps a better approach to create the custom unmarshaler on the `UserID` field itself instead of the whole `UserIdentity`.

This article is a great example: https://engineering.bitnami.com/articles/dealing-with-json-with-non-homogeneous-types-in-go.html

alexkappa:
Closing as this issue has been addressed in another PR.

PR #96, opened 2020-02-19
Add support for passwordless email connection options

dimitrit:
This PR adds support for managing passwordless email templates in Auth0 (which appears to be an undocumented Auth0 Management API feature).  A slight complication is that `options.email` is a `bool` in the case of google-oauth2 strategies, hence the need for a custom unmarshaller.

### DIFF
--- a/management/connection.go
+++ b/management/connection.go
@@ -1,5 +1,7 @@
 package management
 
+import "encoding/json"
+
 type Connection struct {
 	// A generated string identifying the connection.
 	ID *string `json:"id,omitempty"`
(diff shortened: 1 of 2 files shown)


dimitrit:
Have refactored the code so that difference between the different possible types of ConnectionOptionsEmail is made explicit.  This also simplifies marshalling/unmarshalling.

Example:
```
func main() {
	m, err := management.New(domain, id, secret)
	if err != nil {
		panic(err)
	}

	c, err := m.Connection.ReadByName("email")
	if err != nil {
		panic(err)
	}

	var ok bool
	var email *management.PasswordlessEmail
	if email, ok = c.Options.Email.PasswordlessEmail(); !ok {
		panic("not a passwordless email")
	}

	fmt.Printf("old body: %s\n", *email.Body)

	email.Body = auth0.Stringf("content updated at %s", time.Now().String())

	id := *c.ID

	c.ID = nil       // read-only
	c.Name = nil     // read-only
	c.Strategy = nil // read-only
	c.Realms = nil   // read-only

	c.Options.Email.SetPasswordlessEmail(email)

	err = m.Connection.Update(id, c)
	if err != nil {
		panic(err)
	}
}
```

alexkappa:
Hi @dimitrit, thanks for giving this a go. In general, I am a bit reluctant on adding support for things that Auth0 doesn't support officially, but this seems like a reasonable feature to support.

Seeing how this has the extra complication of custom marshal/un-marshal I'll need some time to look into it, and perhaps look into other sdks on how the deal with such problems.

I'll come back to you with some proper comments on this PR.

Cheers,
Alex

dimitrit:
@alexkappa I assume (hope?) this is a documentation oversight rather than it being an unsupported feature.  In any case, my motivation for updating this library is to add support for passwordless emails in the terraform provider (cf. [issue 118](https://github.com/alexkappa/terraform-provider-auth0/issues/118) ).

alexkappa:
Indeed most likely it is so. A quick check inspecting the dashboard, it seems the email, profile, and a slew of google services are listed the same way, used as oauth2 scopes.

What is a bit suspicious is that there is also a scopes array, which sort of works in a similar way (but using urls). It doesn't consistently get set from my admittedly rushed testing. Maybe its supposed to replace the former?

dimitrit:
@alexkappa let me know if you find any issues.  Meanwhile, I have found a slightly more up to date auth0 terraform repository, which I've [forked and updated](https://github.com/dimitrit/terraform-provider-auth0-official).

dimitrit:
@alexkappa any chance of getting this PR merged? Anything I can do to progress this? 
Also, which of the two terraform providers is the 'master'? You appear to be making changes to both [alexkappa/terraform-provider-auth0](https://github.com/alexkappa/terraform-provider-auth0) and [terraform-providers/terraform-provider-auth0](https://github.com/terraform-providers/terraform-provider-auth0). Thanks!

alexkappa:
Hi @dimitrit, I have reached out to Auth0 to seek some guidance and waiting for their response. 

Please be patient while I try to clarify some things. Perhaps you could help me out by looking at other auth0 sdks (preferably statically typed) on how they deal with this situation.

alexkappa:
In [auth0/auth0-java](https://github.com/auth0/auth0-java/blob/master/src/main/java/com/auth0/json/mgmt/Connection.java#L22) the connection options are specified as a `Map<String, Object>` which is equivalent to a `map[string]interface{}` in Go.

Because of how Auth0 just throwing scopes in the options object, I would rather take the same stance and change options to be a `map[string]interface{}`.

What happens on the terraform provider however is a different story, with specific flatteners and expanders per connection type (not great, but I don't see much choice).

As to where to submit your PRs, I'd say keep mine. I'll keep them both up to date as much as I can.

alexkappa:
Hi @dimitrit, I've been thinking about this issue and I prefer to break down the options into specific concrete types per connection type and use a discriminator to unmarshal correctly.

I've written a small prototype to see if this can be done in a reasonable way, and came up with this:
https://gist.github.com/alexkappa/7f868284bcef10da96a40e6ece63698d

There are pros and cons with this approach, but field conflicts should be resolved. Let me know if you have any feedback on this approach.

dimitrit:
Hi @alexkappa, although far from ideal, your suggestion does at least make the different connection options explicit. If you think it's helpful, I can have a go at changing this PR accordingly? It should at least draw out possible issues...

alexkappa:
Could you elaborate on why it's far from ideal? What do you consider some downsides?

dimitrit:
@alexkappa what springs to mind are that creating concrete types is a breaking change - it will mean (significant?) rework in the terraform provider, not to mention other projects that have integrated with this library.  A risk would also be the (lack of) documentation provided by Auth0, which will make it difficult to ensure the concrete types have all the correct fields - it also makes this fragile wrt Auth0 making changes in their api. Perhaps this is why the Java lib uses `Map<String, Object>` 😄 
All that said, I am keen to get to an 'official' auth0 terraform provider that supports our need, so would be happy to try making the changes you suggest.

alexkappa:
@dimitrit that makes sense, thanks for the feedback. And I agree with what you are saying.

> it will mean (significant?) rework in the terraform provider

This seems unavoidable. With the field being the way it is we have to branch off when dealing with `github` or `google` strategies.

> creating concrete types is a breaking change

This is also unavoidable. Fixing this means breaking changes, either way.

Our last resort, and I have conflicting thoughts about this, is `map[string]interface{}`. It can work with any input, and is straightforward to understand and implement. But it leaves all the hard work to users. It's kind of like "you deal with this crap now" solution...

alexkappa:
Another iteration of the polymorphic idea: https://gist.github.com/alexkappa/eb5a7c5f7bd70db3610ed7b4655dc686

Pros:

- This relies on a single field to set options, making it much cleaner imo. 
- It's defined as an `interface{}` but can use concrete type during unmarshal if its available.
- If a concrete type is not available, it falls back to being a `map[string]interface{}`.

Cons:

- Relies on type assertion

alexkappa:
#104

alexkappa:
Closing this as connection options have been heavily reworked in #104 and passwordless has been supported since.

PR #99, opened 2020-02-22
Branding: support page_background with gradient

### LINKED ISSUE #94: Support page_background gradient for Branding
Endpoint: https://auth0.com/docs/api/management/v2#!/Branding/patch_branding

Currently only the string syntax for `colors.page_background` is supported. It would be great, if you could also support the Color Gradient object syntax.

Currently supported syntax:

```json5
{
  "colors": {
    "page_background": "#f9f9f9", // <- Only hex string
    "primary": "#e74c3c"
  },
  "favicon_url": "https://example.com/favicon.ico",
  "logo_url": "https://example.com/logo.png"
}
```

Color Gradient Syntax:

```json5
{
  "colors": {
    "page_background": { // Full object to describe color gradient
      "type": "linear-gradient",
      "start": "#f3c2bd",
      "end": "#e74c3c",
      "angle_deg": 35
    },
    "primary": "#e74c3c"
  },
  "favicon_url": "https://example.com/favicon.ico",
  "logo_url": "https://example.com/logo.png"
}
```

apricote:
Closes #94

### DIFF
--- a/management/branding.go
+++ b/management/branding.go
@@ -1,5 +1,9 @@
 package management
 
+import (
+	"encoding/json"
+)
+
 type Branding struct {
 	// Change password page settings
 	Colors *BrandingColors `json:"colors,omitempty"`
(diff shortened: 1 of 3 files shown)


management/branding.go
```diff
@@ -16,10 +20,102 @@ type Branding struct {
...
+	return json.Marshal(data)
+}
+
+// UnmarshalJSON implements the json.Unmarshaler interface.
+//
+// It is required to handle the json field page_background, which can either
+// be a hex color string, or an object describing a gradient.
+func (bc *BrandingColors) UnmarshalJSON(data []byte) error {
+	// Unmarshal common fields
+	type JSONBrandingColors BrandingColors
```
apricote: By defining a type alias we can unmarshal the common fields (`Primary`) without running into a stack overflow.

management/branding.go
```diff
@@ -16,10 +20,102 @@ type Branding struct {
...
+type jsonPageBackgroundGradient struct {
+	PageBackgroundGradient *BrandingPageBackgroundGradient `json:"page_background,omitempty"`
+}
+
+// MarshalJSON implements the json.Marshaler interface.
+//
+// It is required to handle the json field page_background, which can either
+// be a hex color string, or an object describing a gradient.
+func (bc *BrandingColors) MarshalJSON() ([]byte, error) {
+	var data interface{}
```
apricote: Using `interface{}` and inline struct definitions feels non-idiomatic, do you have any suggestions here @alexkappa ?

apricote:
@alexkappa I added some test for the custom Marshalling/Unmarshalling logic.

apricote:
I rebased and force pushed to fix a conflict in `management/management.gen.go`. I invalidly added some new generated function from a different PR. Looks clean now.

apricote:
@alexkappa Can you take a look at this? I would love to add support for this to the terraform-provider.

alexkappa:
Hi @apricote, sure thing. Sorry for the delay. I'll give it a look today.

management/branding.go
```diff
@@ -16,10 +20,106 @@ type Branding struct {
...
+	bc.Primary = commonBrandingColors.Primary
+
+	// Unmarshal page_background
+	// Can be either a string with solid color
+	var solidBrandingColor jsonPageBackgroundSolid
+	err = json.Unmarshal(data, &solidBrandingColor)
+	if err == nil {
+		bc.PageBackground = solidBrandingColor.PageBackground
+		bc.PageBackgroundGradient = nil
+	}
```
alexkappa, line 110: Instead of trying out which Unmarshal doesn't return errors, we should unmarshal the field in question into an intermediate `interface{}` and from its type we can do what we need to do.

I'd call this field `RawPageBackground` of type `interface{}`, local only to the `MarshalJSON` or `UnmarshalJSON` methods. This would have the appropriate json tag (`json:"page_background,omitempty"`).

With this field populated, an assertion would be sufficient to tell us which kind it is. [This example](https://github.com/go-auth0/auth0/blob/master/management/user.go#L118-L129) does exactly that.

management/branding.go
```diff
@@ -16,10 +20,106 @@ type Branding struct {
...
 	Primary *string `json:"primary,omitempty"`
+
 	// Page background color
+	// Only one of PageBackground and PageBackgroundGradient should be set.
+	// If both fields are set, PageBackground takes priority.
+	PageBackground *string `json:"-"`
+
+	// Page background gradient
+	// Only one of PageBackground and PageBackgroundGradient should be set.
+	// If both fields are set, PageBackground takes priority.
```
alexkappa, line 31: I would go as far as return an error during Marshal/Unmarshal if both are set. What do you think?

management/branding.go
```diff
@@ -16,10 +20,106 @@ type Branding struct {
...
 
+type jsonPageBackgroundGradient struct {
+	PageBackgroundGradient *BrandingPageBackgroundGradient `json:"page_background,omitempty"`
+}
+
+// MarshalJSON implements the json.Marshaler interface.
+//
+// It is required to handle the json field page_background, which can either
+// be a hex color string, or an object describing a gradient.
+func (bc *BrandingColors) MarshalJSON() ([]byte, error) {
```
alexkappa, line 54: Using the approach from `UserIdentity` this could look like so:

```
func (bc *BrandingColors) MarshalJSON() ([]byte, error) {

	type brandingColors BrandingColors
	type brandingColorsWrapper struct {
		*brandingColors
		RawPageBackground interface{} `json:"page_background,omitempty"`
	}

	alias := &brandingColorsWrapper{(*brandingColors)(bc), nil}

	if bc.PageBackground != nil {
		alias.RawPageBackground = bc.PageBackground
	} else if bc.PageBackgroundGradient != nil {
		alias.RawPageBackground = bc.PageBackgroundGradient
	}

	return json.Marshal(alias)
}
```

management/branding.go
apricote: The example looks really good.

After some issues with the type switch and `map[string]interface{}` (I forgot to add the `{}`...), I got this working. To convert from the `map[string]interface{}` to `BrandingPageBackgroundGradient` I first converted back to JSON and then into the correct struct.

management/branding.go
apricote: During Unmarshal this is not an issue, as the JSON can only contain one of both.

For Marshal I added a check to the if-statement to see if both are `!= nil` and return an error in that case.

management/branding.go
apricote: This was resolved by using an embedded type for all normal fields of `BrandingColor`, and an `interface{}` for variable `page_background` field.

See https://github.com/go-auth0/auth0/pull/99#discussion_r403211848

management/branding.go
apricote: That works perfectly. I added a check to see if both fields are `!= nil` and throw an error in that case and added credits to the commit for you.

apricote:
@alexkappa Thanks for your review. I updated my code and re-added the unit tests for the Marshalling/Unmarshalling that got lost when i force-pushed the last time.

alexkappa:
Thank you @apricote, much appreciated! I've merged this in.

PR #102, opened 2020-03-11
Workaround for UserIdentity.UserID being a number

smaant:
Auth0 API has a bug - in some cases it might return `identity.user_id` as a number. Particularly for GitHub identities for requests to the `/users/<user_id` endpoint. In this case JSON unmarshalling will fail with an error: 
`"json: cannot unmarshal number into Go struct field UserIdentity.identities.user_id of type string"`

I created a report in Auth0 community: https://community.auth0.com/t/users-user-id-returns-inconsistent-type-for-identities-user-id/39236
Meanwhile this PR should allow to work around this problem.


alexkappa:
Anton, thank you for working on this. I've made some modification to match it with ConnectionOptions and merged. Cheers!

PR #104, opened 2020-03-13
Make ConnectionOptions "polymorphic"

alexkappa:
Due to the way the Auth0 API handles the various options depending on connection strategy, this SDK is no longer able to continue using a single type for `Connection.Options`.

Depending on the strategy, options fields (and their types) can vary. Sometimes fields are re-used and with different data types. For example `options.email` can be a boolean when strategy is `google-oauth2` and an object when strategy is `email`.

In order to avoid such corner cases, I've introduced specific types for each strategy, and made `Connection.Options` an `interface{}` in order to support any one of the available choices.

The following table shows the type which will be used when Unmarshalling.

| Strategy | Type |
| ------- | --------|
| `auth0` | ConnectionOptions |
| `google-oauth2` | ConnectionOptionsGoogleOAuth2 |
| `facebook` | ConnectionOptionsFacebook |
| `apple` | ConnectionOptionsApple |
| `linkedin` | ConnectionOptionsLinkedin |
| `github` | ConnectionOptionsGitHub |
| `windowslive` | ConnectionOptionsWindowsLive |
| `salesforce` | ConnectionOptionsSalesforce |
| `email` | ConnectionOptionsEmail |
| `sms` | ConnectionOptionsSMS |
| `oidc` | ConnectionOptionsOIDC |
| `waad` | ConnectionOptionsAzureAD |

All other connection strategies can be specified using a `map[string]interface{}`. If you use a strategy that is not included in this list, you are welcome to submit an issue and/or a pull request.

See `ExampleConnectionManager_List` and `ExampleConnectionManager_Create` for sample usage.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -33,4 +33,4 @@ jobs:
       with:
         codeclimate-test-reporter-id: ${{ secrets.CC_TEST_REPORTER_ID }}
         codeclimate-test-reporter-version: latest
-        command: after-build --prefix gopkg.in/auth0.v3
+        command: after-build --prefix gopkg.in/auth0.v4
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Auth0 Go SDK
 
-[![GoDoc](https://godoc.org/gopkg.in/auth0.v3?status.svg)](https://godoc.org/gopkg.in/auth0.v3)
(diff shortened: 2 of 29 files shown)


alexkappa:
Likely this will come with in v4 so we dont break existing dependents.

PR #107, opened 2020-04-02
add "HookSecrets" support

yinzara:
Mind making a tagged release after? :)


management/hook_secrets.go
```diff
@@ -0,0 +1,40 @@
...
+
+type HookSecrets = map[string]string
+
+type HookSecretsManager struct {
+	*Management
+}
+
+func newHookSecretsManager(m *Management) *HookSecretsManager {
+	return &HookSecretsManager{m}
+}
```
alexkappa, line 11: I think we should add this to the `HookManager` instead. See the `UserManager` and `RoleManager` for examples.

management/hook_secrets.go
```diff
@@ -0,0 +1,40 @@
...
+
+// Upserts hook secrets
+//
+// See: https://auth0.com/docs/api/management/v2#!/Hooks/post_secrets
+func (m *HookSecretsManager) Upsert(hookId string, r *HookSecrets) (err error) {
+	return m.post(m.hookPath(hookId), r)
+}
```
alexkappa, line 18: [`POST /api/v2/hooks/{id}/secrets`](https://auth0.com/docs/api/management/v2#!/Hooks/post_secrets) doesn't have upsert semantics. Posting a map that contains an existing field will result in a conflict.

```
{
  "statusCode": 409,
  "error": "Conflict",
  "message": "Error: Secret exists",
  "errorCode": "SecretExists"
}
```

management/hook_secrets.go
```diff
@@ -0,0 +1,40 @@
...
+
+// Reads hook secrets
+//
+// Note: For security, hook secret values cannot be retrieved outside rule
+// execution (they all appear as "_VALUE_NOT_SHOWN_")
+//
+// See: https://auth0.com/docs/api/management/v2#!/Rules_Configs/get_rules_configs
+func (m *HookSecretsManager) Read(hookId string) (r *HookSecrets, err error) {
```
alexkappa, line 26: When moved to the `HookManager` this could be called `Secrets`.

management/hook_secrets.go
```diff
@@ -0,0 +1,40 @@
...
+	err = m.get(m.hookPath(hookId), &r)
+	return
+}
+
+// Delete a list of hook secret keys from a given hook's secrets identified by its hookId and the keys
+//
+// See: https://auth0.com/docs/api/management/v2#!/Rules_Configs/delete_rules_configs_by_key
+func (m *HookSecretsManager) Delete(hookId string, keys ...string) (err error) {
```
alexkappa, line 34: Perhaps we can add a `Keys() []string` method to `HookSecrets` to help with specifying keys?

management/hook_secrets.go
```diff
@@ -0,0 +1,40 @@
...
+	return m.request("DELETE", m.hookPath(hookId), keys)
+}
+
+func (m *HookSecretsManager) hookPath(hookId string) string {
+	return m.uri("hooks", hookId, "secrets")
+}
```
alexkappa, line 40: I'm being a little pedantic here but this is not the hooks path, it's the hooks secrets path. I'd also remove it altogether.

management/hook_secrets_test.go
```diff
@@ -0,0 +1,69 @@
...
+		TriggerID: auth0.String("pre-user-registration"),
+		Enabled:   auth0.Bool(false),
+	}
+
+	err := m.Hook.Create(hook)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	t.Cleanup(func() {
```
alexkappa, line 29: I was not aware of this function! Thanks! TIL.

management/hook_secrets.go
yinzara: OMG It's even worse than that. The PATCH doesn't work if you specify a new secret field.  I found the only way to add a new secret to a hook that already has a secret is to completely DELETE all secrets for that hook then "POST" with all keys and values (even though you can't retrieve the values via the API).  I've got a ticket open with Auth0 support about this.

management/hook_secrets.go
yinzara: I see your point. I'll move it into the hook manager.

management/hook_secrets.go
yinzara: Done

management/hook_secrets.go
yinzara: Done

management/hook_secrets.go
yinzara: I get it. I hate duplicating code but I see how it's kinda already copied and pasted everywhere already. I'll remove it.

management/hook_secrets_test.go
yinzara: I wasn't aware of it until my IDE told me it existed :)

management/hook.go
```diff
@@ -77,3 +80,57 @@ func (m *HookManager) List(opts ...ListOption) (r *HookList, err error) {
...
+
+// Keys gets the configured hook secret keys
+func (s *HookSecrets) Keys() []string {
+	keys := make([]string, len(*s))
+	i := 0
+	for k := range *s {
+		if len(k) > 0 {
+			keys[i] = k
+		}
+		i++
```
alexkappa, line 133: This nil check has the potential to leave the `keys` slice with less elements than it was made to hold right? Why must we check for `len(k) > 0`?

alexkappa: This is the only question remaining from my side, otherwise this PR is good to go

alexkappa:
Thanks Matt!

PR #110, opened 2020-04-30
Add more MFA endpoints

Serjlee:
Because of some custom integration of MFA, we needed a few more Management API endpoints not yet exposed by this library:
- https://auth0.com/docs/api/management/v2#!/Guardian/delete_enrollments_by_id
- https://auth0.com/docs/api/management/v2#!/Users/post_recovery_code_regeneration
- https://auth0.com/docs/api/management/v2#!/Users/get_enrollments

Let me know if I need to tweak something in order to merge!

### DIFF
--- a/management/guardian.go
+++ b/management/guardian.go
@@ -84,6 +84,13 @@ func (m *MultiFactorManager) List(opts ...RequestOption) (mf []*MultiFactor, err
 	return
 }
 
+// Delete an enrollment
+//
+// See: https://auth0.com/docs/api/management/v2#!/Guardian/delete_enrollments_by_id
+func (m *MultiFactorManager) Delete(id string) error {
+	return m.Request("DELETE", m.URI("guardian", "enrollments", id), nil)
+}
+
 type MultiFactorSMS struct{ *Management }
 
 // Enable enables or disables the SMS Multi-factor Authentication.
--- a/management/user.go
+++ b/management/user.go
@@ -397,7 +397,7 @@ func (m *UserManager) RegenerateRecoveryCode(id string) (*UserRecoveryCode, erro
 //
 // See: https://auth0.com/docs/api/management/v2#!/Users/get_enrollments
 func (m *UserManager) Enrollments(id string) (enrols []*UserEnrollment, err error) {
-	err = m.Request("POST", m.URI("users", id, "enrollments"), &enrols)
+	err = m.Request("GET", m.URI("users", id, "enrollments"), &enrols)
 	return enrols, err
 }
 
@@ -405,6 +405,6 @@ func (m *UserManager) Enrollments(id string) (enrols []*UserEnrollment, err erro
 //
 // It's an undocumented API, see: https://community.auth0.com/t/when-will-api-v2-users-userid-authenticators-be-documented/52722
 func (m *UserManager) ListAuthenticators(id string) (enrols []*UserAuthenticator, err error) {
-	err = m.Request("POST", m.URI("users", id, "authenticators"), &enrols)
+	err = m.Request("GET", m.URI("users", id, "authenticators"), &enrols)
 	return enrols, err
 }
--- a/management/user.go
+++ b/management/user.go
@@ -403,7 +403,7 @@ func (m *UserManager) Enrollments(id string) (enrols []*UserEnrollment, err erro
 
 // ListAuthenticators retrieves all authenticators for a user.
 //
-// It's an undocumented API ¯\_(ツ)_/¯
+// It's an undocumented API, see: https://community.auth0.com/t/when-will-api-v2-users-userid-authenticators-be-documented/52722
 func (m *UserManager) ListAuthenticators(id string) (enrols []*UserAuthenticator, err error) {
 	err = m.Request("POST", m.URI("users", id, "authenticators"), &enrols)
 	return enrols, err
--- a/management/user.go
+++ b/management/user.go
@@ -172,6 +172,46 @@ type UserList struct {
 	Users []*User `json:"users"`
 }
 
+// UserRecoveryCode contains the new recovery code for the user
+type UserRecoveryCode struct {
+	// RecoveryCode is the new account recovery code
+	RecoveryCode *string `json:"recovery_code"`
+}
+
+// UserEnrollment contains information about the Guardian enrollments for the user
+type UserEnrollment struct {
+	// Authentication method for this enrollment. Can be `authentication`, `guardian`, or `sms`.
+	AuthMethod *string `json:"auth_method,omitempty"`
+	// Start date and time of this enrollment.
+	EnrolledAt *time.Time `json:"enrolled_at,omitempty"`
+	// ID of this enrollment.
+	ID *string `json:"id,omitempty"`
+	// Device identifier (usually phone identifier) of this enrollment.
+	Identifier *string `json:"identifier,omitempty"`
+	// Last authentication date and time of this enrollment.
+	LastAuth *time.Time `json:"last_auth,omitempty"`
+	// Name of enrollment (usually phone number).
+	Name *string `json:"name,omitempty"`
+	// Phone number for this enrollment.
+	PhoneNumber *string `json:"phone_number,omitempty"`
+	// Status of this enrollment. Can be `pending` or `confirmed`.
+	Status *string `json:"status,omitempty"`
+	// Type of enrollment.
+	Type *string `json:"type,omitempty"`
+}
+
+// UserAuthenticator contains information about an Authenticator
+type UserAuthenticator struct {
+	// ID of this authenticator.
+	ID *string `json:"id,omitempty"`
+	// Type of authenticator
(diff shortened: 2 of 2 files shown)


management/grant.go
```diff
@@ -38,3 +40,12 @@ func (m *GrantManager) List(opts ...ListOption) (g []*Grant, err error) {
...
 }
+
+// Delete revokes all grants associated with a user-id
+// https://auth0.com/docs/api/management/v2#!/Grants/delete_grants_by_id
+func (m *GrantManager) DeleteByUserID(userID string, opts ...ListOption) error {
+	opts = append(opts, func(urls url.Values) {
+		urls.Add("user_id", userID)
+	})
+	return m.delete(m.uri("grants") + m.q(opts))
+}
```
alexkappa, line 51: Deleting by user id should already be possible:

```go
Delete("", Parameter("user_id", userID))
```

... at least that was the intention of allowing `opts ...ListOption` as an argument to `Delete`. Usually deletes only take an id.

Would this work for you?

management/guardian.go
```diff
@@ -84,6 +84,13 @@ func (mfm *MultiFactorManager) List(opts ...ListOption) (mf []*MultiFactor, err
 	return
 }
 
+// Deletes an enrollment
```
alexkappa, line 87:
```suggestion
// Delete an enrollment
```

This should get rid of the warning

management/user.go
```diff
@@ -340,3 +368,20 @@ func (m *UserManager) Blocks(id string) ([]*UserBlock, error) {
...
+	return r, err
+}
+
+// ListEnrollments retrieves all Guardian enrollments for a user.
+//
+// See: https://auth0.com/docs/api/management/v2#!/Users/get_enrollments
+func (m *UserManager) ListEnrollments(id string) (enrols []*UserEnrollment, err error) {
+	err = m.get(m.uri("users", id, "enrollments"), &enrols)
+	return enrols, err
+}
```
alexkappa, line 387:
```suggestion
// Enrollments retrieves all Guardian enrollments for a user.
//
// See: https://auth0.com/docs/api/management/v2#!/Users/get_enrollments
func (m *UserManager) Enrollments(id string) (enrolls []*UserEnrollment, err error) {
	err = m.get(m.uri("users", id, "enrollments"), &enrolls)
	return enrolls, err
}
```

This should make it similar to other sub resource list methods such as `Blocks(...)`, `Permissions(...)`, `Roles(...)` etc.

management/user.go
```diff
@@ -340,3 +368,20 @@ func (m *UserManager) Blocks(id string) ([]*UserBlock, error) {
 func (m *UserManager) Unblock(id string) error {
 	return m.delete(m.uri("user-blocks", id))
 }
+
+// GenerateRecoveryCode removes the current multi-factor authentication recovery code and generate a new one.
+//
+// See: https://auth0.com/docs/api/management/v2#!/Users/post_recovery_code_regeneration
+func (m *UserManager) GenerateRecoveryCode(id string) (*UserRecoveryCode, error) {
```
alexkappa, line 375: Should this be `RegenerateRecoveryCode(...)`? As I understand it, the action is to replace an existing recovery code with a new one.

alexkappa (changes requested):
@Serjlee please let me know if you plan to accommodate the review comments.

Serjlee:
Hey @alexkappa! I'm back from the dead, I'll address your suggestions in a bit

management/grant.go
Serjlee: Yes! I simply missed it.

management/user.go
Serjlee: I think you are right, fixing

Serjlee:
@alexkappa done: I took the chance to squash my changes since I had to rebase to handle the way requests are now built

management/guardian.go
```diff
@@ -84,6 +84,13 @@ func (m *MultiFactorManager) List(opts ...RequestOption) (mf []*MultiFactor, err
 	return
 }
 
+// Delete an enrollment
+//
+// See: https://auth0.com/docs/api/management/v2#!/Guardian/delete_enrollments_by_id
+func (m *MultiFactorManager) Delete(id string) error {
```
lambdalisue, line 90: I think it would be better to add EnrollmentManager here while the API is for an enrollment rather than a factor.

ref: https://github.com/go-auth0/auth0/pull/157/files#diff-e2b523a5aacf304cdcd66ffc8c82c95a9deac4f83af0eb75380fecf23a57a8bfR105

management/user.go
```diff
@@ -343,3 +383,28 @@ func (m *UserManager) Blocks(id string, opts ...RequestOption) ([]*UserBlock, er
...
+//
+// See: https://auth0.com/docs/api/management/v2#!/Users/get_enrollments
+func (m *UserManager) Enrollments(id string) (enrols []*UserEnrollment, err error) {
+	err = m.Request("POST", m.URI("users", id, "enrollments"), &enrols)
+	return enrols, err
+}
+
+// ListAuthenticators retrieves all authenticators for a user.
+//
+// It's an undocumented API ¯\_(ツ)_/¯
```
lambdalisue, line 406: I've asked about that API in Auth0 community so I think you can use the following link at least for now.
https://community.auth0.com/t/when-will-api-v2-users-userid-authenticators-be-documented/52722

lambdalisue (commented):
fyi. Your initial version was for v4 thus I've made quite a similar PR for v5 already at https://github.com/go-auth0/auth0/pull/157

Serjlee:
@lambdalisue introducing an EnrollmentManager sounds like a good idea, but at this point I wouldn't mind closing this PR in favour of yours since they would became kinda identical.

@alexkappa  having either PR merged is ok for me: let us know what to do!

management/user.go
Serjlee: Hah, interesting that you found this PR while looking for ways to manage MFA.

I truly don't understand how that endpoint is still undocumented: it was recommended to me by an Auth0 engineer (via their support) since a while now, and it's a  very valuable endpoint.

lambdalisue:
> @lambdalisue introducing an EnrollmentManager sounds like a good idea, but at this point I wouldn't mind closing this PR in favour of yours since they would became kinda identical.

Note that mine does not include `ListAuthenticators` while it's not documented yet. I use the following code to access that endpoint **in my production code** rather than the library code.

```go
// https://community.auth0.com/t/when-will-api-v2-users-userid-authenticators-be-documented/52722
type authenticator struct {
	ID         *string    `json:"id"`
	Type       *string    `json:"type"`
	Confirmed  *bool      `json:"confirmed"`
	Name       *string    `json:"name,omitempty"`
	CreatedAt  time.Time  `json:"created_at"`
	EnrolledAt *time.Time `json:"enrolled_at,omitempty"`
	LastAuthAt *time.Time `json:"last_auth_at,omitempty"`
	LinkID     *string    `json:"link_id,omitempty"`
}

func getUserAuthenticators(auth0UserID string, opts ...management.RequestOption) (ators []*authenticator, err error) {
	err = m.Request(
		"GET",
		m.URI("users", auth0UserID, "authenticators"),
		&ators,
		opts...,
	)
	return
}
```

Serjlee:
I thought of exposing it in the library after receiving this reassuring reply from Auth0:

> I am not quite sure why it is not documented. However, I confirmed with engineering that if it is safe to use this endpoint. They confirmed that is fine and there are couple of other customers using this endpoint.

(but since in v5 it's possible to build requests from outside the package, I'm fine even without it)

management/user.go
```diff
@@ -343,3 +383,28 @@ func (m *UserManager) Blocks(id string, opts ...RequestOption) ([]*UserBlock, er
...
+func (m *UserManager) Enrollments(id string) (enrols []*UserEnrollment, err error) {
+	err = m.Request("POST", m.URI("users", id, "enrollments"), &enrols)
+	return enrols, err
+}
+
+// ListAuthenticators retrieves all authenticators for a user.
+//
+// It's an undocumented API, see: https://community.auth0.com/t/when-will-api-v2-users-userid-authenticators-be-documented/52722
+func (m *UserManager) ListAuthenticators(id string) (enrols []*UserAuthenticator, err error) {
+	err = m.Request("POST", m.URI("users", id, "authenticators"), &enrols)
```
schan219, line 408: Hello and thanks for making these additions! Just wanted to pop by and make a quick comment. Using POST for the request in `ListAuthenticators` gave me a 404, but using GET works. I think we might need to use GET instead (it's probably the same for `Enrollments` as well).

lambdalisue:
@alexkappa So what should we do next? Close this PR and move to #157 or continue here? Let us know.

management/user.go
Serjlee: Yikes, thanks, That's why committing stuff late at night is a terrible idea.

Serjlee:
@lambdalisue's I'm closing this PR in favour of [yours](https://github.com/go-auth0/auth0/pull/157) since you are definitely keeping a keener eye on this

Issue #77, opened 2019-10-17
Add option to use custom base http.Client


someone1:
There is no way to provide a custom base http.Client to use which may be necessary for things like tracing, proxies, etc.

alexkappa:
This likely relates to #78. If we take google/go-github as a reference implementation they allow both using a custom http.Client and context.Context in each call.

alexkappa:
With #151 merged this is now possible.

PR #112, opened 2020-05-11
Add an option to use custom base http.Client.

rene00:


rene00:
@alexkappa could I get some feedback on this PR which I'm hoping can solve https://github.com/go-auth0/auth0/issues/77?

management/management.go
```diff
@@ -127,18 +127,19 @@ func New(domain, clientID, clientSecret string, options ...apiOption) (*Manageme
...
-	if err != nil {
-		return nil, err
+		_, err = oauth2.Token(m.ctx)
+		if err != nil {
+			return nil, err
+		}
+		m.http = client.New(m.ctx, oauth2)
+		m.http = client.WrapDebug(m.http, m.debug)
+		m.http = client.WrapUserAgent(m.http, m.userAgent)
+		m.http = client.WrapRateLimit(m.http)
```
alexkappa, line 140: I think we should keep these independent of the client, so we can accept `WithHTTPClient` and `WithUserAgent` options simultaneously for example.

So perhaps these are better left outside the `if m.http == nil { ... }` block

management/management.go
```diff
@@ -127,18 +127,19 @@ func New(domain, clientID, clientSecret string, options ...apiOption) (*Manageme
...
+		oauth2 := client.OAuth2(m.url, clientID, clientSecret)
 
-	_, err = oauth2.Token(m.ctx)
-	if err != nil {
-		return nil, err
+		_, err = oauth2.Token(m.ctx)
+		if err != nil {
+			return nil, err
+		}
+		m.http = client.New(m.ctx, oauth2)
```
alexkappa, line 137: I am not sure whether we should completely forego authentication entirely if a custom client has been used. Perhaps we can create new or wrap depending on the options supplied?

If possible, something like this would be nice imo:

```go
if m.http == nil {
	m.http = client.New(m.ctx, m.oauth2)
} else {
	m.http = client.Wrap(m.http, m.ctx, m.oauth2)
}
```

management/management.go
rene00: I had the problem that `client.WrapUserAgent()` would overwrite a users `http.Client.Transport` and any other properties (eg; `http.Client.Timeout`).

I think I've come up with a way which we can wrap `http.Client` without overwriting bits of the users http.Client. Could you let me know how you feel about https://github.com/go-auth0/auth0/pull/112/files#diff-2c846caea57b2b1d6e49c6417f337625R73-R76?

My tests so far are showing that when `client.WrapUserAgent()` is run outside of `if m.http == nil { ... }` and the user supplies a `http.Client`, that the users client is preserved.

I'm now working on figuring out how I can use Transport with `client.WrapDebug()` and then `client.WrapRateLimit()`.

management/management.go
rene00: >  Perhaps we can create new or wrap depending on the options supplied?

Good idea. I'll come up with something for review.

> If possible, something like this would be nice imo:
> 
> ```go
> if m.http == nil {
> 	m.http = client.New(m.ctx, m.oauth2)
> } else {
> 	m.http = client.Wrap(m.http, m.ctx, m.oauth2)
> }
> ```
Great, thanks for the example. 🙇

alexkappa:
@rene00 I've just [discovered something](https://pkg.go.dev/golang.org/x/oauth2?tab=doc#example-Config-CustomHTTP) that I believe you might find useful 😉 

It seems the oauth2 package allows setting a custom `http.Client` using `context.Context`. this way we can define our `Wrap` function as follows:

```go
func Wrap(c *http.Client, ctx context.Context, cc *clientcredentials.Config) *http.Client {
	return New(context.WithValue(ctx, oauth2.HTTPClient, c), cc)
}
```

Hope this helps!

internal/client/client.go
```diff
@@ -92,6 +92,13 @@ func WrapDebug(c *http.Client, debug bool) *http.Client {
 	return c
 }
 
+func WrapHTTPClientWithOauth2(ctx context.Context, httpClient *http.Client, clientCredentials *clientcredentials.Config) *http.Client {
+	ctx = context.WithValue(oauth2.NoContext, oauth2.HTTPClient, httpClient)
+	client := oauth2.NewClient(ctx, clientCredentials.TokenSource(ctx))
+	httpClient.Transport = client.Transport
```
alexkappa, line 98: I am curious why we would need to use the httpClient after we have instructed `oauth2.NewClient` to use it as its base.. Isn't it enough to just return `oauth2.NewClient(...)`?

rene00: I've pushed a minor update so that it uses [clientCredentials.Client()](https://github.com/golang/oauth2/blob/master/clientcredentials/clientcredentials.go#L66-L68) (which will call `oauth2.NewClient()`) but the question is still valid.

From what I understand from reading [oauth2.NewClient()](https://github.com/golang/oauth2/blob/master/oauth2.go#L342-L352) and investigating the example provided in the previous [comment](https://github.com/go-auth0/auth0/pull/112#issuecomment-628082781), it will overwrite the other users supplied `http.Client` fields like `http.Client.Timeout`. I was unclear if this is common practice so I opted for preserving these supplied fields just in case. To help illustrate further:
```go
client := http.Client{Timeout: Timeout: 2 * time.Second}
newClient := client.WrapHTTPClientWithOauth2(ctx, client, clientCredentials)
```
`newClient.Timeout` will be `nil` but the user had it set.

If I've misunderstood this, let me know. Also let me know if you feel that  `WrapHTTPClientWithOauth2()` should overwrite a user supplied field like `http.Client.Timeout` or preserve it. This is also relevant for the other Wrap functions, which attempt to preserve these fields too.

rene00:
@alexkappa I believe this PR is ready for a proper review. Let me know what you think. I'm happy to make further changes based on your input. I'll also add comments once you're good with the changes. Thanks,

internal/client/client.go
```diff
@@ -70,17 +90,16 @@ func WrapDebug(c *http.Client, debug bool) *http.Client {
...
+	return c
+}
+
+// WrapHTTPClientWithOauth2 is used to supply oauth2 support to a user provided http.Client.
+func WrapHTTPClientWithOauth2(ctx context.Context, httpClient *http.Client, clientCredentials *clientcredentials.Config) *http.Client {
+	ctx = context.WithValue(oauth2.NoContext, oauth2.HTTPClient, httpClient)
+	client := clientCredentials.Client(ctx)
+	httpClient.Transport = client.Transport
+	return httpClient
 }
```
alexkappa, line 103: I'd prefer we let the returned client be the one that `oauth2.NewClient()` returns. We may not be able to verify timeouts and such, but it feels cleaner here imo. Do you see any other issues with returning the oauth2 client instead?

This suggestion relates to my comment on `management.New()`, making it accept functional options:

```suggestion
// Client an existing http.Client using the provided OAuth2 configuration.
func Client(ctx context.Context, c *http.Client, cfg *clientcredentials.Config, options ...func(*http.Client)) *http.Client {
	for _, option := range options {
		option(c)
	}
	ctx = context.WithValue(ctx, oauth2.HTTPClient, c)
	return oauth2.NewClient(ctx, cfg.TokenSource(ctx))
}
```

internal/client/client.go
```diff
@@ -70,17 +90,16 @@ func WrapDebug(c *http.Client, debug bool) *http.Client {
...
 
 func New(ctx context.Context, c *clientcredentials.Config) *http.Client {
```
alexkappa, line 105: We can remove this

internal/client/client.go
```diff
@@ -20,40 +20,60 @@ import (
...
-type RoundTripFunc func(*http.Request) (*http.Response, error)
+func newTransport(base http.RoundTripper, headers map[string]string, debug bool) *Transport {
+	if base == nil {
+		base = http.DefaultTransport
+	}
+	return &Transport{base, headers, debug}
+}
 
-func (rf RoundTripFunc) RoundTrip(req *http.Request) (*http.Response, error) {
-	return rf(req)
```
alexkappa, line 56: I'd prefer to keep the `RoundTripFunc` instead of `Transport`.

internal/client/client.go
```diff
@@ -20,40 +20,60 @@ import (
...
+func WrapUserAgent(c *http.Client, userAgent string) *http.Client {
+	c.Transport = newTransport(c.Transport, map[string]string{"User-Agent": userAgent}, false)
+	return c
 }
```
alexkappa, line 77: To support the functional arguments passed into `Client(...)` we'll have to change the signature (and name) to allow this:

```go
func WithRateLimit() func(*http.Client) { ... }
```

```suggestion
// WithUserAgent wraps c.Transport in a way that adds a User-Agent header to
// each request.
func WithUserAgent(userAgent string) func(*http.Client) {
	return func(c *http.Client) {
		baseTransport := c.Transport
		c.Transport = RoundTripFunc(func(req *http.Request) (*http.Response, error) {
			req.Header.Set("User-Agent", userAgent)
			return baseTransport.RoundTrip(req)
		})
	}
}
```

The same should happen for `WrapDebug` and `WrapRateLimit`.

srenatus:
🤔 This is a good improvement, it's a shame to let it go to waste. Anyone listening? Anything I can do to help with this?

rene00:
> 🤔 This is a good improvement, it's a shame to let it go to waste. Anyone listening? Anything I can do to help with this?

Hi 👋 I'm listening ✋ I've been meaning to get back to this PR pending free cycles. Please free to submit a new PR though.

alexkappa:
`WithClient` was introduced as an option with the release of v5. Closing this one.

Issue #114, opened 2020-05-19
DisplayName field missing in Connection struct


juniocezar:
The connection struct in `management/connection.go` doesn't have a field for getting or setting the Display Name for Auth0 connections.

Basically, the change needed is to change this:

```go
type Connection struct {
	// A generated string identifying the connection.
	ID *string `json:"id,omitempty"`

	// The name of the connection. Must start and end with an alphanumeric
	// character and can only contain alphanumeric characters and '-'. Max
	// length 128.
	Name *string `json:"name,omitempty"`
        ...
```

into this:

```go
type Connection struct {
	// A generated string identifying the connection.
	ID *string `json:"id,omitempty"`

	// The name of the connection. Must start and end with an alphanumeric
	// character and can only contain alphanumeric characters and '-'. Max
	// length 128.
	Name *string `json:"name,omitempty"`
	DisplayName *string `json:"display_name,omitempty"`
        ....
```

ref: https://github.com/go-auth0/auth0/blob/4ac1f71b27731a3313eb9d0addd237775b1a16c9/management/connection.go#L36

PR #115, opened 2020-05-21
Add DisplayName field to the Connection struct

juniocezar:
Auth0 Connections have an optional field Display Name, which is missing from the Connection struct.

image

fixes: https://github.com/go-auth0/auth0/issues/114

### DIFF
--- a/management/connection.go
+++ b/management/connection.go
@@ -33,7 +33,8 @@ type Connection struct {
 	// The name of the connection. Must start and end with an alphanumeric
 	// character and can only contain alphanumeric characters and '-'. Max
 	// length 128.
-	Name *string `json:"name,omitempty"`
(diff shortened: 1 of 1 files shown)


management/connection.go
```diff
@@ -34,6 +34,7 @@ type Connection struct {
 	// character and can only contain alphanumeric characters and '-'. Max
 	// length 128.
 	Name *string `json:"name,omitempty"`
+	DisplayName *string `json:"display_name,omitempty"`
```
alexkappa, line 37: Could you please run `go fmt` on this file?

juniocezar: @alexkappa done!

alexkappa:
Thanks @juniocezar, I've merged it in and will tag it soon. Stay tuned!

PR #123, opened 2020-07-09
Updated Code Scanning Workflow

MAHDTech:
This PR updates the Code Scanning Workflow to include a step which uploads the results. Any findings are then available from the security tab within the repository.


alexkappa:
Huh, thanks for doing this @MAHDTech, I feel a little embarrassed I missed this :)

PR #124, opened 2020-07-18
Ianthpun/handleconveriblebool

ianthpun:
Create a new `ConvertibleBoolean` type that handles both string and bool type marshal/unmarshalling.

Comes with a bucket of tests to prove it works


custom.go
```diff
@@ -0,0 +1,48 @@
+package auth0
+
+import "strconv"
+
+// ConvertibleBoolean is a custom bool type that is able to unmarshal/marshal from both string and bool types
+type ConvertibleBoolean bool
```
ianthpun, line 6: Im not married to having this stuff inside `package auth0`, so if you have any better ideas, lemme know (maybe `internal/custom`)?

PR #138, opened 2020-09-15
add missing SAML Connection Options

yvovandoorn:
Three SAML options got missed when SAML Connection support was re-added to the provider. 

This PR brings parity to what settings can be set with regards to the SAML connection.


jbauth0 (approved):
Lgtm :shipit:

alexkappa (commented):
Hi Yvo, thanks for adding the new fields! Could you re-generate getters by running go generate ./...?

PR #141, opened 2020-09-17
Add support for Oauth2 Connection Strategy

### LINKED ISSUE #139: Support for Oauth2 Connection
It seems that Oauth2 connection strategy is not implemented yet?
We would like to use terraform to manage our Custom Social Connectors using Oauth2.
See the terrform issue here: https://github.com/alexkappa/terraform-provider-auth0/issues/266

mcalster:
Fixes #139

### DIFF
--- a/management/connection.go
+++ b/management/connection.go
@@ -22,7 +22,7 @@ const (
 	ConnectionStrategyEmail               = "email"
 	ConnectionStrategySMS                 = "sms"
 	ConnectionStrategyOIDC                = "oidc"
-	ConnectionStrategyOauth2              = "oauth2"
+	ConnectionStrategyOAuth2              = "oauth2"
 	ConnectionStrategyAD                  = "ad"
 	ConnectionStrategyAzureAD             = "waad"
 	ConnectionStrategySAML                = "samlp"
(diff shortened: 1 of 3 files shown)


mcalster:
@alexkappa I think this is good to go know. Can you have a look.

management/connection.go
```diff
@@ -526,6 +529,41 @@ func (c *ConnectionOptionsOIDC) SetScopes(enable bool, scopes ...string) {
...
+type ConnectionOptionsOAuth2 struct {
+	ClientID              *string `json:"client_id,omitempty"`
+	ClientSecret          *string `json:"client_secret,omitempty"`
+	AuthorizationEndpoint *string `json:"authorization_endpoint"`
+	TokenEndpoint         *string `json:"token_endpoint"`
+	Scope                 *string `json:"scope,omitempty"`
+
+	// Scripts for the connection
+	// Allowed keys are: "fetchUserProfile"
+	CustomScripts map[string]interface{} `json:"scripts,omitempty"`
```
alexkappa, line 541: Why did you go for a `CustomScripts` here instead of `Scripts` as in the JSON tag?

alexkappa (commented):
Hi @mcalster thanks for working on this! I have only one question but your PR looks aligned well with the other connection options (particularly OIDC which is very similar).

management/connection.go
mcalster: I chose this approach to keep the terraform schema as it is in the other cases. I think it would be nice to keep the same name for the same functionality, and then just change the json to satisfy the API.  
I can change it if you want

alexkappa: Yeah, I see what you mean... But I think I would prefer to remain closer to the Auth0 API naming and make no assumptions on our end.

mcalster: I've changed to follow the API names.

management/connection_test.go
```diff
@@ -181,7 +181,7 @@ func TestConnectionOptions(t *testing.T) {
 
 	t.Run("OAUTH2", func(t *testing.T) {
```
alexkappa, line 182:
```suggestion
	t.Run("OAuth2", func(t *testing.T) {
```

PR #142, opened 2020-09-22
management: add WithContext method

srenatus:
This tries to introduce the ability to provide a context, while keeping the API
surface intact, and not changing every single method signature.

It is modelled slightly after how gorp had done the same thing three years ago:
https://github.com/go-gorp/gorp/pull/352

### DIFF
--- a/management/management.go
+++ b/management/management.go
@@ -139,6 +139,11 @@ func New(domain, clientID, clientSecret string, options ...apiOption) (*Manageme
 	m.http = client.WrapUserAgent(m.http, m.userAgent)
 	m.http = client.WrapRateLimit(m.http)
 
+	m.initManagers()
+	return m, nil
+}
+
+func (m *Management) initManagers() {
 	m.Client = newClientManager(m)
 	m.ClientGrant = newClientGrantManager(m)
 	m.Connection = newConnectionManager(m)
@@ -161,8 +166,6 @@ func New(domain, clientID, clientSecret string, options ...apiOption) (*Manageme
 	m.Guardian = newGuardianManager(m)
 	m.Prompt = newPromptManager(m)
 	m.Blacklist = newBlacklistManager(m)
-
-	return m, nil
 }
 
 func (m *Management) uri(path ...string) string {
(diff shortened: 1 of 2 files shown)


alexkappa:
Hi @srenatus, isn't [`management.WithContext()`](https://github.com/go-auth0/auth0/blob/master/management/management.go#L278-L284) sufficient for your use case? You can use it as an argument to `management.New(...)`.

srenatus:
Hi!

I may be mistaken, but wouldn't that acquire a new token for every call to
new? I think I'd rather avoid that...

alexkappa:
> Hi! I may be mistaken, but wouldn't that acquire a new token for every call to new? I think I'd rather avoid that...

Why would you need to make multiple calls to New?

srenatus:
I'd like to pass a different context in for each request. The existing
means would have me call new for every call, wouldn't they?

srenatus:
Heya. I've been having limited access to a keyboard yesterday, so I'd like to describe my use case some more now:

Imagine there's some HTTP handler code, "LIST auth thingies", and that code ends up calling `mgmt.Client.List()`, where `mgmt` is an instance of `management.Management` shared among the different handlers. When the requesting party goes away, I'd like the _list_ request to be cancelled. I can imagine some ways of making this happen:

1. Instantiate a separate instance of `management.Management` in every handler, use `management.New(..., management.WithContext(ctx))`.
2. Share the instance, use `mgmt.WithContext(ctx).Client.List()`.
3. Alter every method, allowing us to do `mgmt.Client.List(ctx)`

(1.) comes with the disadvantage that ever call to `New()` ends up retrieving a new access token using the client creds flow, work that wouldn't be necessary.
I'd think that (3.) is the cleanest way to do it, but also the most effort. So, (2.) seems like a decent compromise. What do you think?

(There's more alternatives, to explore, I suppose. If `New()` would take in a token source, we would perhaps be able to avoid the token business, and still have an instance per handler invocation.... 🤔)

alexkappa:
Thanks for elaborating on the use case, I understand it's a reasonable ask to pass a different context with each request.

There is also the approach taken by other sdks such as [google/go-github](/google/go-github) where each function takes a `ctx` as argument. For example `mgmt.Client.List(ctx, ...)`, which might more idiomatic and allows for separate ctx to be passed with each call.

I admit that the `WithContext(ctx)` fluent style method is an interesting idea. As you noted won't impact the existing method signatures. My gripe with it is that there is a lot of copying around the managers, which I must admit is due to the way the code is structured currently.

Can I think about this a little more? Sorry this is taking a while but its a feature I want to consider well before accepting in.

alexkappa:
@srenatus in the v5 branch I've reworked `ListOptions` into `RequestOptions` and accept them in all methods. These options include the ability to pass your own context to each method.

Example:
```go
mgmt.Client.List(management.Context(ctx))
mgmt.Client.Read("xyz", management.Context(ctx))
mgmt.Client.Update("xyz", c, management.Context(ctx))
```

Thanks for bringing this issue up. I think the API is more usable with this change.

srenatus:
Thanks, this is a good improvement! 👍

PR #143, opened 2020-09-24
internal/client: simplify funs passed to rehttp.NewTransport

srenatus:
I believe them to be equivalent... :)


alexkappa:
Very nice simplification @srenatus! Thanks for this :)

PR #147, opened 2020-09-28
fixed wrong property names for tokenURL and authorizationURL

### LINKED ISSUE #146: Bug in Oauth2 Connection Support
While the request generated by the current implementation of the Oauth2 connection strategy, the token_endpoint and authorization_endpoint are not used by the Oauth2 Social connection.

The correct names are tokenURL and authorizationURL.

mcalster:
Fixes #146 
Blocks: https://github.com/alexkappa/terraform-provider-auth0/pull/272

### DIFF
--- a/management/connection.go
+++ b/management/connection.go
@@ -530,11 +530,11 @@ func (c *ConnectionOptionsOIDC) SetScopes(enable bool, scopes ...string) {
 }
 
 type ConnectionOptionsOAuth2 struct {
-	ClientID              *string `json:"client_id,omitempty"`
-	ClientSecret          *string `json:"client_secret,omitempty"`
-	AuthorizationEndpoint *string `json:"authorization_endpoint"`
-	TokenEndpoint         *string `json:"token_endpoint"`
-	Scope                 *string `json:"scope,omitempty"`
+	ClientID         *string `json:"client_id,omitempty"`
+	ClientSecret     *string `json:"client_secret,omitempty"`
+	AuthorizationURL *string `json:"authorizationURL"`
+	TokenURL         *string `json:"tokenURL"`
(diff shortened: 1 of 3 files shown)


mcalster:
@alexkappa This is ready to go. I've introduced two new attributes following the API naming.
Please review.

alexkappa:
Hi @mcalster is this documented somewhere? Perhaps we should include these fields in our test case so we catch this earlier.

mcalster:
@alexkappa 
> is this documented somewhere?
I was only able to find it by creating the oauth2 connection, and then extract it via the API. Here is an example response:
    {
        "id": "con_12345",
        "options": {
            "scope": "openid",
            "scripts": {
                "fetchUserProfile": "async function retrieveUserProfile(accessToken, context, callback) { callback(null, null);\n}"
            },
            "**tokenURL**": "https://example.com/oauth2/token",
            "client_id": "test-client",
            "client_secret": "SoooSecret",
            "**authorizationURL**": "https:/example.com/oauth2/authorize"
        },
        "strategy": "oauth2",
        "name": "oauth2-test",
        "is_domain_connection": false,
        "realms": [
            "oauth2-test"
        ],
        "enabled_clients": []
    },
I haven't been able to find it in the API documentation.
 >Perhaps we should include these fields in our test case so we catch this earlier.
It's is hard to verify before hand? Since all four properties are accepted as valid parameters for an oauth2 connection.

alexkappa (approved):

mcalster (commented):
LGTM

mcalster:
@alexkappa anything I can do to push this forward? I know this is a spare time project, so if I can help in anyway :-)

PR #151, opened 2020-10-08
V5 (alternative proposal)

alexkappa:
### Request and RequestOption

This PR enables support for passing custom `context.Context` to individual requests. Instead of introducing a new argument to all methods, the optional final argument is re-worked to allow passing a context, or any other manipulation to an `http.Request`.

This has a subtle change to the API but a much less invasive one - adding a `context.Context` argument to all calls.

```diff
- api.Client.Read("xxx")
+ api.Client.Read("xxx", Context(ctx))
```

To achieve this, the `ListOption` type was repurposed into what is now `RequestOption`. This type can be optionally present as the last argument with all methods of the SDK.

### New and ManagementOption

The `New` constructor has also been revisited to allow alternative authentication mechanisms, such as static authentication tokens. The standard client credentials grant authentication mechanism is now configurable using the `WithClientCredentials` option.

```diff
- m, err = New(domain, id, secret)
+ m, err = New(domain, WithClientCredentials(id, secret))
```

A number of additional options have been defined, such as passing a custom HTTP client (see `WithClient`), authenticating using a static token (see `WithStaticToken`). The type of these options has been renamed to `ManagementOption` to avoid confusion between `RequestOption`.

### Request, NewRequest, and Do

Several methods have now been exposed to enable arbitrary requests to the Auth0 Management API. The `Request` method should support most such cases. For more control, `NewRequest`, `Do` and `URI` are also available.


lambdalisue:
I really like this new version thanks! :tada: Now I can finally use `context.Context` on individual requests 👍

Issue #127, opened 2020-07-19
Change Tenant SessionLifetime and IdleSessionLifetime to float64


alexkappa:
Change data type of `Tenant` `SessionLifetime` and `IdleSessionLifetime` to `float64`.

It seems that the Auth0 API allows saving fractions of an hour in these fields causing the marshaling/unmarshaling to fail when such values are presented.

```diff
type Tenant struct {
 
        // Login session lifetime, how long the session will stay valid (unit:
        // hours).
-       SessionLifetime *int `json:"session_lifetime,omitempty"`
+       SessionLifetime *float64 `json:"session_lifetime,omitempty"`
 
        // Force a user to login after they have been inactive for the specified number (unit: hours)
-       IdleSessionLifetime *int `json:"idle_session_lifetime,omitempty"`
+       IdleSessionLifetime *float64 `json:"idle_session_lifetime,omitempty"`
```

See: https://github.com/alexkappa/terraform-provider-auth0/issues/233

bergbrains:
@alexkappa, what's the path forward for implementing this change?  We are very interested in moving this forward.

PR #156, opened 2020-11-02
Send tenant session_lifetime_in_minutes when session_lifetime is < 1

alexkappa:
`Tenant` `session_lifetime` and `idle_session_lifetime` are changed to `float64` types.

Also, when `session_lifetime` is less than 1, it is replaced with `session_lifetime_in_minutes` instead, multiplied by 60.

Fixes: #127

### DIFF
--- a/auth0.go
+++ b/auth0.go
@@ -31,6 +31,20 @@ func IntValue(i *int) int {
 	return 0
 }
 
+// Float64 returns a pointer to the float64 value passed in.
+func Float64(f float64) *float64 {
+	return &f
+}
+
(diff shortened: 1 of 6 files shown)


jbauth0 (approved):
Looks like a good way to deal with the API quirkiness.

alexkappa:
@jbauth0 I've made some additional changes to convert to minutes _only_ if timeout is < 1 hour. Otherwise values are rounded to the nearest integer and kept as is.

bergbrains:
Great to see this one moving forward and thank for your work on this, @alexkappa.  When can we expect this to be released?

PR #160, opened 2020-11-09
Updated Metadata property on connection to be string map

robotsrule:
First off, I have no experience in GoLang so I'm not sure this is the correct solution to this problem.  Please let me know if anything needs to be changed with this PR.

This is step one of me attempting to fix an issue with the terraform provider that utilizes this library. 
[Issue 282 on alexkappa/terraform-provider-auth0](https://github.com/alexkappa/terraform-provider-auth0/issues/282)

When attempting to fix that in the terraform provider I noticed that the data type for the property was an interface pointer which doesn't make sense to me.

I think the Metadata property on connection should be the same data type as "client_metadata" would be on an application/client since connection metadata functions in exactly the same way as client metadata.  I changed it to map[string]string instead of *interface{} and modified a test case to set the property and expect that it was set.

### DIFF
--- a/management/connection.go
+++ b/management/connection.go
@@ -70,7 +70,7 @@ type Connection struct {
 	// connection name will be added as realm.
 	Realms []interface{} `json:"realms,omitempty"`
 
-	Metadata *interface{} `json:"metadata,omitempty"`
+	Metadata map[string]string `json:"metadata,omitempty"`
 }
 
 func (c *Connection) MarshalJSON() ([]byte, error) {
--- a/management/connection_test.go
+++ b/management/connection_test.go
@@ -308,10 +308,6 @@ func TestConnectionOptions(t *testing.T) {
 		g := &Connection{
 			Name:     auth0.Stringf("Test-SAML-Connection-%d", time.Now().Unix()),
 			Strategy: auth0.String("samlp"),
-			Metadata: map[string]string{
-				"foo": "bar",
-				"bat": "baz",
-			},
 			Options: &ConnectionOptionsSAML{
 				SignInEndpoint: auth0.String("https://saml.identity/provider"),
 				SigningCert: auth0.String(`-----BEGIN CERTIFICATE-----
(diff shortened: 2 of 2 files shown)


management/connection.go
```diff
@@ -70,7 +70,7 @@ type Connection struct {
 	// connection name will be added as realm.
 	Realms []interface{} `json:"realms,omitempty"`
 
-	Metadata *interface{} `json:"metadata,omitempty"`
+	Metadata map[string]string `json:"metadata,omitempty"`
```
paddycarey, line 73: I think this should be `map[string]interface{}`, to allow values other than strings.

management/connection_test.go
```diff
@@ -354,5 +358,7 @@ yE+vPxsiUkvQHdO2fojCkY8jg70jxM+gu59tPDNbw3Uh/2Ij310FgTHsnGQMyA==
 
 		expect.Expect(t, o.GetSignInEndpoint(), "https://saml.identity/provider")
 		expect.Expect(t, o.GetTenantDomain(), "example.com")
+		expect.Expect(t, g.Metadata["foo"], "bar")
+		expect.Expect(t, g.Metadata["bat"], "baz")
```
paddycarey, line 362: I'm not sure this test adds much value here, since it's just checking that the values set initially are still set. If we really wanted to check that setting metadata worked we should set it, then retrieve a new instance via a `GET` request to check against.

I'm not sure it's necessary though, since we're not doing the same for other fields, only on the `Options` field currently.

paddycarey (commented):
Hey @robotsrule thanks for the contribution! This looks good, I've left a few comments but nothing too major, let me know what you think :)

paddycarey:
This closes https://github.com/go-auth0/auth0/issues/179

management/connection.go
robotsrule: @paddycarey thanks for the feedback.

I just experimentally verified against Auth0's management API that it will only accept string values.  I tried setting objects and numbers and it failed validation and returned a 400 in each case.  

Auth0 hasn't done a good job of documenting this particular field but as far as I can determine it will only allow string values.

It is also worth noting that the data type for ClientMetadata on the client resource is also defined as map[string]string in this SDK and not map[string]interface{} 

So I think map[string]string is probably correct in this specific case

management/connection_test.go
robotsrule: I think you are right, I should have looked more closely at what this test is actually doing.  I will remove it.  

The ClientMetadata field on the client resource is similarly untested and that is the most similar thing in the SDK to the connection metadata so it is probably fine.

Issue #158, opened 2020-11-04
Universal Login Template


gagalago:
[Universal login template](https://auth0.com/docs/universal-login/new-experience/universal-login-page-templates#page-templates-api) is missing from this SDK (see [anounce](https://support.auth0.com/notifications/5f526b537b9bed000bed935b))

I can try to work on that but do you have some inside on how to do it before I start?

alexkappa:
Hi @gagalago, thank you for raising this issue. Feel free to give it a try, and let me know if you need any help.

yvovandoorn:
Closing this issue as #161 was merged in last year.

PR #161, opened 2020-11-10
add universal login template

gagalago:
add [universal login template](https://auth0.com/docs/universal-login/new-experience/universal-login-page-templates)

fix https://github.com/go-auth0/auth0/issues/158

### DIFF
--- a/management/branding.go
+++ b/management/branding.go
@@ -116,6 +116,10 @@ type BrandingFont struct {
 	URL *string `json:"url,omitempty"`
 }
 
+type BrandingTemplateUniversalLogin struct {
+	Body *string `json:"body,omitempty"`
+}
+
 type BrandingManager struct {
 	*Management
 }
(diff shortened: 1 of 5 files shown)


gagalago:
@alexkappa you can review this pull request

management/branding.go
```diff
@@ -116,6 +116,10 @@ type BrandingFont struct {
 	URL *string `json:"url,omitempty"`
 }
 
+type BrandingTemplateUniversalLogin struct {
```
gagalago, line 119: @alexkappa I am wondering if I need to create a new standalone resource or if I follow the [documentation](https://auth0.com/docs/api/management/v2#!/Branding/get_universal_login) and add it like it is right now?

gagalago: I saw that you did already the same in user for example

relu (approved):
LGTM :+1:

management/management.go
```diff
@@ -257,6 +257,23 @@ func (m *Management) NewRequest(method, uri string, payload interface{}, options
 	return
 }
 
+// NewHtmlRequest returns a new HTTP request.
+func (m *Management) NewHtmlRequest(method, uri string, html string, options ...RequestOption) (r *http.Request, err error) {
```
alexkappa, line 261: We can probably avoid duplicating this method.

alexkappa (commented):
Hi @gagalago, thanks for working in this. I've made some changes and merged this in.

Issue #164, opened 2020-11-19
Listing connections with strategy but without options will return a json error


rene00:
When listing connections using `management.WithFields("strategy")` the connection options will fail to unmarshal and a `unexpected end of JSON input` error will be returned.

Consider:
```go
 _, err := m.Connection.List(management.WithFields("id", "strategy"))
fmt.Println(err)

_, err = m.Connection.List(management.WithFields("id", "strategy", "options"))
fmt.Println(err)

 _, err = m.Connection.List(management.WithFields("id", "name"))
fmt.Println(err)
```
which prints:
```
unexpected end of JSON input
<nil>
<nil>
```

I believe the error happens within [UnmarshalJSON](https://github.com/go-auth0/auth0/blob/ec73ed30bc5b8ae0e17fc87f70b25123f6c8feb2/management/connection.go#L97-L162):
```
		err = json.Unmarshal(w.RawOptions, &v)
		if err != nil {
			return err
		}
```
`w.RawOptions` will be `nil` when a returned connection does not contain the `options` field but does contain a `strategy` field.

A more verbose reproduction:
```
% cat main.go
package main

import (
        "encoding/json"
        "fmt"

        "gopkg.in/auth0.v5/management"
)

func main() {
        var data = `{"id":"test","strategy":"auth0","is_domain_connection":false}`
        c := management.Connection{}
        err := json.Unmarshal([]byte(data), &c)
        if err != nil {
                fmt.Printf("err1:%s\n", err)
        }

        data = `{"id":"test","strategy":"auth0","is_domain_connection":false,"options":{"passwordPolicy": "good"}}`
        err = json.Unmarshal([]byte(data), &c)
        if err != nil {
                fmt.Printf("err2:%s\n", err)
        }
}
% go run main.go
err1:unexpected end of JSON input
%
```

https://github.com/go-auth0/auth0/pull/165 should fix this.

rene00:
https://github.com/go-auth0/auth0/pull/165 may fix this.

PR #165, opened 2020-11-19
Only unmarshal connection options if not nil.

rene00:
Fixes https://github.com/go-auth0/auth0/issues/164


alexkappa:
Thanks for submitting this @rene00!

PR #174, opened 2020-12-10
WIP: User.email_verified is union of string|bool

cyx:
Similar to UserIdentity, this is a shim which we can't necessarily fix
without accidentally breaking userspace.

This changes the semantics such that we can handle a string `"true"` or
`"false"` for these boolean types.

### DIFF
--- a/management/user.go
+++ b/management/user.go
@@ -5,6 +5,8 @@ import (
 	"reflect"
 	"strconv"
 	"time"
+
+	"gopkg.in/auth0.v5"
 )
 
 // User represents an Auth0 user resource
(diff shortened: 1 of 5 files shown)


management/user.go
```diff
@@ -62,7 +64,7 @@ type User struct {
 	// True if the user's email is verified, false otherwise. If it is true then
 	// the user will not receive a verification email, unless verify_email: true
 	// was specified.
-	EmailVerified *bool `json:"email_verified,omitempty"`
+	EmailVerified *auth0.BoolString `json:"email_verified,omitempty"`
```
alexkappa, line 67: It's best to keep this field as a `*bool` and handle the discrepancy inside the marshal/unmarshal methods.

Similar to [`UserIdentity`](https://github.com/go-auth0/auth0/blob/master/management/user.go#L113), we can define the tag as `json:"-"` and using a `RawEmailVerified` within an alias and wrapper during [unmarshal](https://github.com/go-auth0/auth0/blob/master/management/user.go#L127-L131).

There is no need to create a new type `auth0.BoolString` and we probably want to avoid confusing people with it.

alexkappa:
Work continued at #181

PR #178, opened 2020-12-24
Add post_identities endpoint support

jmaeso:
Opening PR since looks to be the only way of running tests. Pending to verify if the feature works

### DIFF
--- a/management/user.go
+++ b/management/user.go
@@ -2,6 +2,7 @@ package management
 
 import (
 	"encoding/json"
+	"net/http"
 	"reflect"
 	"strconv"
 	"time"
@@ -98,9 +99,13 @@ type UserIdentityLink struct {
 	// Connection id of the secondary user account being linked when more than one auth0 database provider exists.
 	ConnectionID *string `json:"connection_id,omitempty"`
 	// Secondary account user id.
-	UserID string `json:"user_id"`
+	UserID *string `json:"user_id,omitempty"`
 	// Identity provider of the secondary user account being linked.
-	Provider string `json:"provider"`
+	Provider *string `json:"provider,omitempty"`
+	// LinkWith requires the authenticated primary account's JWT in the Authorization header.
+	// Must be a JWT for the secondary account being linked. If sending this parameter,
+	// provider, user_id, and connection_id must not be sent.
+	LinkWith *string `json:"link_with,omitempty"`
 }
 
 type UserIdentity struct {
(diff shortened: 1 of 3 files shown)


jmaeso:
Sorry, I'm not sure how to run the Test action 😅

management/user.go
```diff
@@ -343,3 +343,19 @@ func (m *UserManager) Blocks(id string, opts ...RequestOption) ([]*UserBlock, er
...
+//
+// See: https://auth0.com/docs/api/management/v2#!/Users/post_identities
+func (m *UserManager) Link(id string, il *IdentityLink, opts ...RequestOption) (err error) {
+	return m.Request("POST", m.URI("users", id, "identities"), il, opts...)
+}
+
+// IdentityLink contains the data needed for linking an identity to a given user.
+//
+// Note: connection_id param is needed just to identify a particular database connection for the 'auth0' provider.
+type IdentityLink struct {
```
alexkappa, line 357: Perhaps we should use the `User...` prefix to maintain the convention.

management/user.go
```diff
@@ -343,3 +343,19 @@ func (m *UserManager) Blocks(id string, opts ...RequestOption) ([]*UserBlock, er
...
+	ConnectionID *string `json:"connection_id,omitempty"`
+	UserID       string  `json:"user_id"`
+	Provider     string  `json:"provider"`
```
alexkappa, line 360: Can we include the comments from the api docs?

```
// Identity provider of the secondary user account being linked.
Provider *string ...
```

`UserID` and `Provider` should also be pointers.

We're missing `link_with` according to the api docs: JWT for the secondary account being linked. If sending this parameter, provider, user_id, and connection_id must not be sent.

management/user.go
alexkappa: Also maybe move the type definition above the methods to be in line with the rest of the code.

management/user_test.go
```diff
@@ -260,6 +260,37 @@ func TestUser(t *testing.T) {
 		}
 		t.Logf("%v\n", us)
 	})
+
+	t.Run("Link", func(t *testing.T) {
+		ulAlice, err := m.User.Search(Query(`email:"alice@example.com"`))
```
alexkappa, line 265: We can also create a couple users, that better represent linking between two identities. We can make a better example of linking Batman to Bruce Wayne or Heisenberg to Walter White 😄 

Just remember to clean up once done

```go
t.Cleanup(func() { 
	m.User.Delete(walter.GetID())
	m.User.Delete(heisenberg.GetID())
})
```

alexkappa (commented):
Hi @jmaeso, thank you for taking the time to put this together! 

The code looks pretty good. I made a couple of suggestions on aligning it more with the conventions used throughout the codebase.

Once ready you can ping me again so I can test/merge.

Cheers,
Alex

jmaeso:
Good for the comments, they make sense, I'll address them ASAP.
This morning though I tried to test the implementation. Unfortunately doesn't work, it will need some changes/additions into the internals of the library:
The way the `manager.Request()` works is that reuses the request's body to unmarhsal the server's response and in this case, both the payloads are different and cannot be reused.
Not sure how would you address this point. Playing with`RequestOption` (not sure if would work)? A new `Request` method? Update the existent method to decouple request & response types?

management/user.go
jmaeso: For the `link_with` param, it actually requires a JWT from the primary account in the `authorization` header. I think this library does not support this specific authentication mechanism, that's why I didn't add the param; Not adding the param made the `UserID` and `Provider` params required (so non-pointers).
WDYT?

jmaeso:
ping @alexkappa
Last changes are tested on some personal script, so should pass the tests. This being said, if you have concerns about the chosen solution, feel free to discuss alternatives

management/user.go
alexkappa: `UserID` and `Provider` should be pointers anyway as to preserve the conventions used throughout the SDK.

alexkappa: I'd also include `link_with` just in case. Perhaps someone might want to use it together with other tools?

alexkappa:
Thanks @jmaeso, sorry about the request/response limitations you are experiencing here. It's quite uncommon but happens from time to time the payload of the request being different than that of the response.

For now I would suggest calling the contents of `RequestWithCustomResponse` directly within the `Link` method. I can try to think of a way to make parsing the response easier in the future.

jmaeso:
Comments applied @alexkappa

alexkappa:
Nicely done @jmaeso! I've tested locally and merged it in. Will publish a release shortly.

Issue #148, opened 2020-09-29
Cannot clear fields in PATCH request


sebastian-stephan:
This is also related to https://github.com/go-auth0/auth0/issues/74.

The resource structs use pointers, which will never be marshalled when set to `nil` due to the `omitempty` tag. However sometimes that would be required in a PATCH request if you want to clear a field.

As an example the `given_name` field on a user with the [Patch User](https://auth0.com/docs/api/management/v2#!/Users/patch_users_by_id) request: Auth0 does not accept an empty string, and the correct way to clear this field would be to send `null` in the JSON payload.

I briefly looked for inspiration how this is solved in other SDKs, and found a similar bug in the GitHub SDK: https://github.com/google/go-github/issues/236 where the suggested solution is to manually create a request.

I guess to do this one could expose the underlying `*http.Client` with a getter or make the `request` function public. Then you could construct the payload manually:
```golang
payload := struct {
    GivenName *string `json:"given_name"`
}{}
```
Not sure what the other options would be. I guess one could always replace all pointers with some custom type?

alexkappa:
Hi @sebastian-stephan, with the changes made in #151 it is now possible to gain access to the underlying client in order to make the request manually.

The `NewRequest`, `Do` and `Request` methods are probably what you're looking for. Also the `URI` method can help with specifying the right URL path.

PR #193, opened 2021-02-03
Fixed problem with omitted enabled_clients on connection

apamildner:
### Proposed Changes
Not omitting empty `enabled_clients`, as this does not work as expected.
* Related to https://github.com/alexkappa/terraform-provider-auth0/issues/325

### Community Note

### DIFF
--- a/management/connection.go
+++ b/management/connection.go
@@ -61,9 +61,8 @@ type Connection struct {
 	Options interface{} `json:"-"`
 
 	// The identifiers of the clients for which the connection is to be
-	// enabled. If the array is empty or the property is not specified, no
-	// clients are enabled.
-	EnabledClients []interface{} `json:"enabled_clients,omitempty"`
+	// enabled. If the array is empty, no clients are enabled.
+	EnabledClients []interface{} `json:"enabled_clients"`
 
 	// Defines the realms for which the connection will be used (ie: email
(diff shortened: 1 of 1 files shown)


paddycarey:
Hey @apamildner, thanks for the contribution!

I think this one requires further consideration, it's very similar to https://github.com/go-auth0/auth0/pull/43 in that it causes unintended consequences when making `PATCH` requests with a partially complete connection.

It's not ideal but there are some workarounds available, such as using the underlying request methods @alexkappa mentions in this issue https://github.com/go-auth0/auth0/issues/148

It's a pretty common use-case though, so we'll need to address it somehow, as I understand it's much more difficult to solve when using the terraform provider than if using the SDK directly.

apamildner:
Ah I see. That other issue is a very similar problem indeed. However, the PATCH request with a partial connection is only working right now because the Auth0 api doesn't work as documented. The api [docs](https://auth0.com/docs/api/management/v2#!/Connections/patch_connections_by_id) states:
image

paddycarey:
@apamildner thanks for pointing this out, I believe that's a docs issue, rather than an implementation one. I'll raise a ticket with the right team to get that clarified.

paddycarey:
@apamildner docs should now be updated, thanks for pointing that out!

PR #201, opened 2021-03-02
add ProvisioningTicketUrl to Connection

seiro-ogasawara:
### Proposed Changes

* Add ProvisioningTicketUrl to property of management.Connection.

This item is not listed in the API documentation, but it is issued in some strategies such as "ad".
```
{
  "id":"con_xxxx",
  "options":{
    "certAuth":false,
    "disable_cache":false,
    "kerberos":false,
    "domain_aliases":[
      "idp-domain.com"
    ],
    "tenant_domain":"xyz-domain.com",
    "ips":null,
    "brute_force_protection":true
  },
  "strategy":"ad",
  "name":"xyz-test-ad-conn",
  "provisioning_ticket_url":"https://xyz-dev.us.auth0.com/p/ad/xxxx",
  "is_domain_connection":false,
  "show_as_button":false,
  "display_name":"Xyz-Test-Ad-Conn",
  "enabled_clients":[],
  "realms":[
    "xyz-test-ad-conn"
  ]
}
```

### Community Note

### DIFF
--- a/management/connection.go
+++ b/management/connection.go
@@ -71,6 +71,9 @@ type Connection struct {
 	Realms []interface{} `json:"realms,omitempty"`
 
 	Metadata map[string]string `json:"metadata,omitempty"`
+
+	// Provisioning Ticket URL is Ticket URL for Active Directory/LDAP, etc.
+	ProvisioningTicketUrl *string `json:"provisioning_ticket_url,omitempty"`
 }
 
 func (c *Connection) MarshalJSON() ([]byte, error) {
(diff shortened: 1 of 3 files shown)


yvovandoorn:
Hi @seiro-ogasawara! Thank you so much for your contribution. 

The use around `"provisioning_ticket_url"` is a bit special. The attribute is generated and managed on the Auth0 server side.  The management API definitely doesn't like it when its part of the payload. 

What would really help here is to write a test that validates the AD connection strategy (see https://github.com/go-auth0/auth0/blob/master/management/connection_test.go) given its a one off and only really used with AD connections. The test would then create, update and delete the object. This validates that as long as the `provisioning_ticket_url` isn't touched nor updated, that changes will work.

seiro-ogasawara:
Thanks for reviewing @yvovandoorn !

I've added a test, how do you think?
I've also `rebase` and `push -f` to resolve conflicts with the latest code, so the commit hash has changed, but I hope you don't mind.

yvovandoorn:
Hey @seiro-ogasawara! That is a rad test 👍.

Not only did you improve the test coverage, I like the update without and with the value. 

Merging!

PR #205, opened 2021-03-23
Have Grant List also use applyListDefaults.

### LINKED ISSUE #168: Enforcement of paginated API responses from the auth0 management v2 API
Hey @alexkappa!

Happy to see you're still maintaining this 👍

We saw an email about paginated responses from the 21st January 2021 affecting these endpoints:
```bash
    GET /api/v2/clients
    GET /api/v2/client-grants
    GET /api/v2/grants
    GET /api/v2/connections
    GET /api/v2/device-credentials (NOTE: pagination will not be enforced when type is not provided)
    GET /api/v2/resource-servers
    GET /api/v2/rules
```

I looked through the source code here and found the [client-grants](https://github.com/go-auth0/auth0/blob/cd598ba2e393c9361df6c82d3a5732505ecafc56/management/client_grant.go) and [resource-servers](https://github.com/go-auth0/auth0/blob/cd598ba2e393c9361df6c82d3a5732505ecafc56/management/resource_server.go) endpoints had support for pagination but not the others.

I thought to create the issue so others might know and update their terraform providers if we did end up patching this

yvovandoorn:
### Proposed Changes

* Have Grant List also use applyListDefaults.

Fixes #168 

#### Acceptance Test Output

```
go test ./... -v -run TestGrant                                     
=== RUN   TestGrant
=== RUN   TestGrant/List
    grant_test.go:15: []
--- PASS: TestGrant (0.47s)
    --- PASS: TestGrant/List (0.47s)
PASS
ok  	gopkg.in/auth0.v5/management	(cached)
```

### Community Note

### DIFF
--- a/management/grant.go
+++ b/management/grant.go
@@ -17,6 +17,11 @@ type Grant struct {
 	Scope []interface{} `json:"scope,omitempty"`
 }
 
+type GrantList struct {
+	List
+	Grants []*Grant `json:"grants"`
+}
+
 type GrantManager struct {
 	*Management
 }
@@ -28,7 +33,7 @@ func newGrantManager(m *Management) *GrantManager {
 // List the grants associated with your account.
 //
 // See: https://auth0.com/docs/api/management/v2#!/Grants/get_grants
-func (m *GrantManager) List(opts ...RequestOption) (g []*Grant, err error) {
+func (m *GrantManager) List(opts ...RequestOption) (g *GrantList, err error) {
 	err = m.Request("GET", m.URI("grants"), &g, applyListDefaults(opts))
 	return
 }
--- a/management/grant.go
+++ b/management/grant.go
@@ -29,7 +29,7 @@ func newGrantManager(m *Management) *GrantManager {
 //
 // See: https://auth0.com/docs/api/management/v2#!/Grants/get_grants
 func (m *GrantManager) List(opts ...RequestOption) (g []*Grant, err error) {
-	err = m.Request("GET", m.URI("grants"), &g, opts...)
+	err = m.Request("GET", m.URI("grants"), &g, applyListDefaults(opts))
 	return
 }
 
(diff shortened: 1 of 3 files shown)


yvovandoorn:
@paddycarey or @jbauth0 a solid double check would be good here. I think this is the behavior we want out of `user_grants`.

management/grant.go
```diff
@@ -29,7 +29,7 @@ func newGrantManager(m *Management) *GrantManager {
 //
 // See: https://auth0.com/docs/api/management/v2#!/Grants/get_grants
 func (m *GrantManager) List(opts ...RequestOption) (g []*Grant, err error) {
-	err = m.Request("GET", m.URI("grants"), &g, opts...)
+	err = m.Request("GET", m.URI("grants"), &g, applyListDefaults(opts))
```
alexkappa, line 32: `applyListDefaults` will add the `include_totals=true` query parameter to the request. It's needed for pagination. The response will differ in that case from an array to an object containing -among other fields- a `grands` array.

```json
{
  "total": 0,
  "start": 0,
  "limit": 50,
  "grants": []
}
```

This means we'll need to create a `GrantList` struct to return here instead of `[]*Grant`.

```go
type GrantList struct {
	List
	Grants []*Grant `json:"grants"`
}
```

management/grant_test.go
```diff
@@ -9,7 +9,7 @@ func TestGrant(t *testing.T) {
 	t.Run("List", func(t *testing.T) {
 		var gs []*Grant
 		gs, err = m.Grant.List()
-		if err != nil {
+		if err == nil {
```
alexkappa, line 12: This test now verifies that when we call `List()` an error should be expected. Are you sure this is right?

management/grant.go
yvovandoorn: Ah yes! That makes sense. Added.

management/grant_test.go
yvovandoorn: You're right with List, then the old situation is true. Reverted.

yvovandoorn:
@alexkappa addressed your comments. Thanks for pointing me to GrantList. 

Should be good to go now!

alexkappa (approved):
Looks great! Thank you @yvovandoorn

PR #207, opened 2021-03-25
ability to create MFA enrollment ticket

nishanths:
This adds support for https://auth0.com/docs/api/management/v2#!/Guardian/post_ticket.

### DIFF
--- a/management/guardian.go
+++ b/management/guardian.go
@@ -1,6 +1,10 @@
 package management
 
-import "time"
+import (
+	"encoding/json"
+	"net/http"
+	"time"
+)
 
 type Enrollment struct {
 	// ID for this enrollment
@@ -33,7 +37,7 @@ type MultiFactor struct {
 type MultiFactorPolicies []string
 
 type MultiFactorProvider struct {
-	//One of auth0|twilio|phone-message-hook
+	// One of auth0|twilio|phone-message-hook
 	Provider *string `json:"provider,omitempty"`
 }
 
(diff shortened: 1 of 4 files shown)


management/guardian.go
```diff
@@ -102,6 +106,53 @@ type EnrollmentManager struct {
...
+		return EnrollmentTicket{}, err
+	}
+
+	res, err := m.Do(req)
+	if err != nil {
+		return EnrollmentTicket{}, err
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusOK {
```
nishanths, line 147: The API doc mentions `204`, but I tried with curl and the response status code was actually a `200`.

nishanths: Please also fix this in the API docs, if you can!

yvovandoorn: Reported internally about the mismatch.

yvovandoorn:
Hi @nishanths! First of all, thank you for your contribution!

Would it be possible to breakout the `CreateTicket` struct out of the function? Since we're dealing with a ticket creation here, you can take a peek at `management/ticket.go`. While your approach is valid, it would be a unique approach in the SDK as a whole.

nishanths:
Thanks for the review. Please take a look at the new changes.

nishanths:
My commit messages on the branch are really bad (e.g., `.`), so consider use the squash option in the Merge button dropdown when you merge.

yvovandoorn:
Hi @nishanths - could you enable the "Allow edits from maintainers" (https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork) option?

Thanks!

nishanths:
I have been unable to find the checkbox to do that in the UI for some reason? Maybe it's because I'm not an "Owner" in the organization in which this pull request's branch exists.

If you can tell me the reason why you'd like the ability to make changes, I can try to make the changes myself.

nishanths:
If it's regarding the commits on the branch, I've tried to clean it up now.

nishanths:
I can also reopen the PR from a repository in which I can allow edits from maintainers. Just let me know. :)

yvovandoorn:
@nishanths For some reason I still can't modify it. 

What I wanted to do was the following:
* On L130, in the func, wrap the function around error instead of doing error handling in the function. Similiar to what it does on L154 (retrieving enrollments) and L162 (deleting enrollments). Sp removing L132 until L149.
* Discard your .gitignore as that should be a separate PR 
* Testing for me is failing as you're missing username, I was going to see if I could fix up the test.

That said, if you want to incorporate the above changes, then I would greatly appreciate it.

nishanths:
> For some reason I still can't modify it.

Sorry about that. I still have been unable to find this option in the UI for this PR. (I can try making a new PR to see if I can choose to allow edits at time of PR creation.) In the meantime, I've made the changes you requested (except the one regarding . Thanks for the feedback.

> Discard your .gitignore as that should be a separate PR

Done.

> Testing for me is failing as you're missing username, I was going to see if I could fix up the test.

Done. I had omitted it initially because the Auth0 tenant I was testing with required a username to not be set. I've added it now.

> On L130, in the func, wrap the function around error instead of doing error handling in the function. Similiar to what it does on L154 (retrieving enrollments) and L162 (deleting enrollments). Sp removing L132 until L149.

Do you mean use `m.Request()`? I don't think that will be possible because we have different request and response types, but `Request()` uses the same type for the request and response?

nishanths:
> I still have been unable to find this option in the UI for this PR.

The reason the allow-maintainer-edits option doesn't exist for this PR, is because the PR branch is from an organization repo, not a user repo. 

https://github.community/t/how-can-we-enable-allow-edits-from-maintainers-by-default/2847/5

yvovandoorn:
> Do you mean use m.Request()? If I understand correctly, I don't think that will be possible because we have different request and response types, but m.Request() requires the same type for the request and response?

Ah yes! Duh. 

Ok in that case LGTM. Shipping.

nishanths:
Thank you for merging, @yvovandoorn. :)

PR #214, opened 2021-05-11
Add support for Anomaly endpoints

Widcket:
### Proposed Changes

This PR adds support for the two Anomaly endpoints:

- `GET	/api/v2/anomaly/blocks/ips/{id}`: https://auth0.com/docs/api/management/v2#!/Anomaly/get_ips_by_id
- `DELETE	/api/v2/anomaly/blocks/ips/{id}`: https://auth0.com/docs/api/management/v2#!/Anomaly/delete_ips_by_id

The GET endpoint, unlike the other API2 endpoints, takes no data in the request body and returns no data in the response body (except in the case of error). The only input is an IP address passed as part of the path, and the outcome of the operation is signaled by status codes.
A status code of `200` means that the IP is blocked, and a status code of `404` means that it is not blocked.
The current request-making logic does not expose the status code, and returning the error does not make sense as a `404`  will end up as an error, when it's a non-error output. So instead of calling `m.Request` directly, the implementation of `CheckIP` adapts the code of `m.Request` to work around these peculiarities.

The DELETE endpoint does not return data in the response body either (except in the case of error), but it does not use an error status code for a non-error output. So `UnblockIP` just calls `m.Request` directly.

#### Acceptance Test Output

This was tested manually by calling a command on a modified Auth0 CLI.

Screen Shot 2021-05-10 at 20 52 32
Screen Shot 2021-05-10 at 21 04 48

### Community Note

### DIFF
--- a/management/anomaly.go
+++ b/management/anomaly.go
@@ -1,5 +1,9 @@
 package management
 
+import (
+	"net/http"
+)
+
 type AnomalyManager struct {
 	*Management
 }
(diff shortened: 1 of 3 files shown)


cyx (approved):
Looks good to me!

Widcket:
@cyx thanks for the review. Can you please merge it? I don't have permissions.

paddycarey (approved):
LGTM

